PR #2, opened 2023-04-26
Nontrivial chunking

aliddell:
Depends on [acquire-core-libs/#2](https://github.com/acquire-project/acquire-core-libs/pull/2) and [acquire-driver-common/#4](https://github.com/acquire-project/acquire-driver-common/pull/4).


andy-sweet (commented):
I'm still working my way through this, but thought it's worth dropping the comments/questions I have so far. 

The existing documentation is great for high level usage. But it could also be very helpful to provide some high level developer docs on the program flow and threading approach used here.

src/chunk.writer.cpp
```diff
@@ -0,0 +1,380 @@
...
+        if (auto frame = writer->pop_frame_and_make_current()) {
+            uint8_t* region = nullptr;
+            while (!roi.finished()) {
+                size_t nbytes = frame->next_contiguous_region(roi, &region);
+                if (0 < nbytes) {
+                    CHECK(nullptr != region);
+                    writer->write(region, region + nbytes);
+                }
+                if (nbytes < bytes_per_row) {
+                    std::fill_n(fill.data(), bytes_per_row - nbytes, 0);
```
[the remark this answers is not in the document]
aliddell: That... is a great point I hadn't thought of. Part of a hasty rewrite around the encoder behavior and I missed that.

src/chunk.writer.hh
```diff
@@ -0,0 +1,72 @@
...
+struct ChunkWriter final
+{
+  public:
+    ChunkWriter() = delete;
+    ChunkWriter(BaseEncoder* encoder,
+                const ImageShape& image,
+                const TileShape& tile,
+                uint32_t tile_col,
+                uint32_t tile_row,
+                uint32_t tile_plane,
```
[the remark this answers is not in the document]
nclack: oh! col, row, plane are indices. Maybe a comment or variable name change would've made it less confusing on first glance.

README.md
```diff
@@ -15,5 +13,67 @@ Compression is done via [blosc][].
...
+There are 4 parameters you can set to determine the chunk size, namely `tile_width`, `tile_height`, `tile_planes`,
+and `bytes_per_chunk`:
+
+```c
+int
+storage_properties_set_chunking_props(struct StorageProperties* out,
+                                      uint32_t tile_width,
+                                      uint32_t tile_height,
+                                      uint32_t tile_planes,
+                                      uint32_t max_bytes_per_chunk)
```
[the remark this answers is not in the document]
aliddell: Went back to core-libs and fixed it ([PR](https://github.com/acquire-project/acquire-core-libs/pull/16)).

src/chunk.writer.hh
```diff
@@ -0,0 +1,74 @@
...
+                const std::string& base_directory);
+    ~ChunkWriter();
+
+    [[nodiscard]] bool write_frame(const TiledFrame& frame);
+
+    std::mutex& mutex() noexcept;
+
+    const uint32_t tile_col;
+    const uint32_t tile_row;
+    const uint32_t tile_plane;
```
nclack, line 45: These aren't accessed outside of this class, so they should be private.

src/chunk.writer.hh
```diff
@@ -0,0 +1,74 @@
...
+
+  private:
+    BaseEncoder* const encoder_{};
```
nclack, line 48: Avoid initialization here.

```suggestion
    BaseEncoder* const encoder_;
```

src/chunk.writer.hh
```diff
@@ -0,0 +1,74 @@
...
+
+    size_t bytes_per_chunk_;
+    size_t tiles_per_chunk_;
+    size_t bytes_written_;
+
+    std::string base_dir_;
+    int current_chunk_;
+    char dimension_separator_;
+    struct file* current_file_;
```
nclack, line 57: It's a little odd to use a pointer here. It looks like you're using the pointer to give `optional` semantics - consider just using `std::optional<struct file>`.

src/tiled.frame.hh
```diff
@@ -0,0 +1,61 @@
...
+#include "prelude.h"
+
+namespace acquire::sink::zarr {
+struct TileShape
+{
+    struct tile_shape_dims_s
+    {
+        uint32_t width, height, planes;
+    } dims;
+};
```
nclack, line 23:
```suggestion
struct TileShape
{
        uint32_t width, height, planes;
};
```

src/tiled.frame.hh
```diff
@@ -0,0 +1,61 @@
...
+    TiledFrame(const VideoFrame* frame,
+               const ImageShape&,
+               const TileShape& tile_shape);
+    TiledFrame(const TiledFrame&) = delete;
+    ~TiledFrame();
+
+    size_t bytes_of_image() const;
+
+    uint64_t frame_id() const;
+    uint8_t* data() const;
```
nclack, line 38: `frame_id()` and `data()` should be made private. `bytes_of_image()` is used in a unit test but otherwise isn't used outside the class. Is there a way to make it private?

src/tiled.frame.hh
```diff
@@ -0,0 +1,61 @@
...
+                                   uint32_t tile_plane) const;
+
+  private:
+    size_t bytes_of_image_;
+    uint64_t frame_id_;
+    uint8_t* buf_;
+    ImageShape image_shape_;
+    TileShape tile_shape_;
+
+    [[nodiscard]] size_t get_contiguous_region(uint8_t** region,
```
nclack, line 53: Needs a comment. Also, comment should address `nodiscard`.

src/tiled.frame.hh
```diff
@@ -0,0 +1,61 @@
...
+               const TileShape& tile_shape);
+    TiledFrame(const TiledFrame&) = delete;
+    ~TiledFrame();
+
+    size_t bytes_of_image() const;
+
+    uint64_t frame_id() const;
+    uint8_t* data() const;
+
+    [[nodiscard]] size_t copy_tile(uint8_t* tile,
```
nclack, line 40: Needs a comment. Also, comment should address `nodiscard`.

nclack (commented):
Added more comments.

src/tiled.frame.hh
aliddell: `frame_id()` and `data()` don't actually appear to be used at all. I guess they're vestigial, so I'll get rid of em til they're needed. Guess I'll do the same with `bytes_of_image()`.

src/zarr.encoder.hh
```diff
@@ -0,0 +1,41 @@
...
+    void allocate_buffer(size_t buf_size);
+
+    virtual void set_file(struct file* file_handle);
+
+  protected:
+    std::vector<uint8_t> buf_;
+    size_t cursor_;
+    size_t bytes_per_pixel_;
+    std::string path_;
+    struct file* file_; // non-owning
```
nclack, line 34: I don't think this merit's a change, but the need for this file_ pointer here is awkward. It's a consequence of using inheritance as a composition mechanism here.

It looks like the owner of the file is always a child class. This guarantees the lifetime of the pointer except in the destructor - the parent gets destroyed last - but things look safe there at the moment.

Anyway, no action here just wanted to raise it just in case.

nclack (commented):
Almost there. Maybe we can finish the review off tomorrow morning.

src/zarr.encoder.hh
aliddell: The `ChunkWriter` owns the file in a `std::optional` wrapper. Either the file is open, so it [has a value](https://github.com/aliddell/acquire-driver-zarr/blob/nontrivial-chunking/src/chunk.writer.cpp#L193) and the Encoder gets a pointer to it, or the file is closed so it [doesn't have a value](https://github.com/aliddell/acquire-driver-zarr/blob/nontrivial-chunking/src/chunk.writer.cpp#L215) and the pointer is null.

tests/CMakeLists.txt
```diff
@@ -36,6 +39,8 @@ else()
         )
 
         add_test(NAME test-${tgt} COMMAND ${tgt})
+        set_tests_properties(test-${tgt} PROPERTIES LABELS anyplatform)
+        set_tests_properties(test-${tgt} PROPERTIES LABELS zarr)
```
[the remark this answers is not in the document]
aliddell: If I run `ctest -L anyplatform` I end up running tests for core-libs and driver-common. Granted these should already be passing and so shouldn't be merged in, but it's not the responsibility of the Zarr driver repo to ensure that upstream tests are passing. You can still run these tests, but if you're just interested in Zarr driver behavior, that's what this label is for.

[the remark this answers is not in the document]
aliddell: But on closer inspection, this doesn't do what I want it to do, it actually clobbers the "anyplatform" label. I just pushed a fix for that.

src/zarr.blosc.hh
```diff
@@ -1,140 +1,67 @@
...
-#include <thread>
-#include <blosc.h>
+#include "zarr.encoder.hh"
+
+#include "blosc.h"
 
 namespace acquire::sink::zarr {
 
-enum class CodecId
+struct BloscCompressor
```
nclack, line 12: maybe rename to something like "compression parameters"

nclack (approved):
Little things.

src/zarr.hh
```diff
@@ -6,21 +6,40 @@
...
+    std::mutex mutex;
+    std::condition_variable cv;
+    bool should_stop;
+};
+
+struct ThreadJob
+{
+    std::shared_ptr<TiledFrame> frame;
+    std::function<bool(std::shared_ptr<TiledFrame>)> f;
+};
```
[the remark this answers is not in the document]
aliddell: Actually this struct isn't used any longer in favor of a full closure. Just going to delete it.

PR #74, opened 2023-07-21
Fix striping in Zarr

### LINKED ISSUE #73: Striping in Zarr
I noticed some striping when acquiring to Zarr from an Orca Fusion. 

PXL_20230718_191617402 MP

I haven't observed this on the latest PyPI release, but only when acquiring from the Orca Fusion (but not the simulated cameras) using main in this repo. I suspect this is due to the fact that simulated cameras do not support multi-byte sample types.

aliddell:
closes #73 

### Fixed

- A bug where multibyte samples exhibited striping behavior due to being copied from the wrong offset in the source
  buffer.

## How this was tested

In `tests/no-striping.cpp` I used a simulated camera to generate radial sin data in u8, u16, and f32 variants (see https://github.com/acquire-project/acquire-driver-common/pull/31 which this PR depends on). (Note that only u16 version of the test has been committed.) I visually verified that while the u8 data displayed correctly in napari, the u16 variant exhibited some striping, and the f32 variant exhibited even more. After making the fix, I verified that each of the u8, u16, and f32 variants displayed correctly in napari.

## Depends on

- [x] https://github.com/acquire-project/acquire-driver-common/pull/31

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,0 +1,50 @@
+if (${WITH_EXAMPLES})
+    #
+    # PARAMETERS
+    #
+    set(project acquire-driver-zarr) # CMAKE_PROJECT_NAME gets overridden if this is a subtree of another project
+
+    #
+    # Examples
+    #
+    set(examples
+            no-striping
+    )
+
+    foreach (name ${examples})
+        set(tgt "${project}-${name}")
+        add_executable(${tgt} ${name}.cpp)
+        target_compile_definitions(${tgt} PUBLIC "EXAMPLE=\"${tgt}\"")
+        set_target_properties(${tgt} PROPERTIES
+                MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>"
+        )
+        target_include_directories(${tgt} PRIVATE "${CMAKE_CURRENT_LIST_DIR}/../")
+        target_link_libraries(${tgt}
+                acquire-core-logger
+                acquire-core-platform
+                acquire-video-runtime
+        )
+    endforeach ()
+
+    #
+    # Copy driver to examples directory
+    #
+    list(GET examples 0 onename)
+
+    foreach (driver
+            acquire-driver-common
+            acquire-driver-zarr
+    )
+        add_custom_target(${project}-copy-${driver}-for-examples
+                COMMAND ${CMAKE_COMMAND} -E copy
+                $<TARGET_FILE:${driver}>
+                $<TARGET_FILE_DIR:${project}-${onename}>
+                DEPENDS ${driver}
+                COMMENT "Copying ${driver} to $<TARGET_FILE_DIR:${project}-${onename}>"
+        )
+
+        foreach (name ${examples})
+            add_dependencies(${tgt} ${project}-copy-${driver}-for-examples)
+        endforeach ()
+    endforeach ()
+endif ()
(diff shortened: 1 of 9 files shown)


tests/no-striping.cpp
```diff
@@ -0,0 +1,232 @@
...
+        }
+        {
+            uint32_t n = consumed_bytes(beg, end);
+            OK(acquire_unmap_read(runtime, 0, n));
+            if (n)
+                LOG("stream %d consumed bytes %d", 0, n);
+        }
+
+        CHECK(nframes == props.video[0].max_frame_count);
+    }
```
nclack, line 179: Do we need to check frames here? Maybe cut. Alternatively, could factor this block into a `check_frame_count()` function.

```suggestion
```

tests/no-striping.cpp
```diff
@@ -0,0 +1,232 @@
...
+    ASSERT_EQ(int, "%d", frame_width, shape[3]);
+
+    auto chunks = zarray["chunks"];
+    ASSERT_EQ(int, "%d", expected_frames_per_chunk, chunks[0]);
+    ASSERT_EQ(int, "%d", 1, chunks[1]);
+    ASSERT_EQ(int, "%d", frame_height, chunks[2]);
+    ASSERT_EQ(int, "%d", frame_width, chunks[3]);
+
+    // check chunked data
+    auto chunk_size = chunks[0].get<int>() * chunks[1].get<int>() *
```
nclack, line 222: trying to make sure I understand the logic behind the test - if the shape and chunk metadata the striping is fixed?

tests/no-striping.cpp
```diff
@@ -0,0 +1,232 @@
+#include "device/hal/device.manager.h"
```
nclack, line 1: For future context, can you add a comment at the top describing what this test asserts? Maybe add a link to the pr.

nclack (commented):
Could you add some clarifying comments to the test? I just want to make sure it's easy to understand later.

nclack (approved):

tests/no-striping.cpp
aliddell: Honestly, the test itself doesn't demonstrate anything. It generates data for a visual inspection. Without knowing `t` in the [expression](https://github.com/acquire-project/acquire-driver-common/blob/main/src/simcams/imfill.pattern.cpp#L54)

```c
127.0f * (sinf(6.28f * (mid * 10.0f + (dx2 + dy2) * 1e-2f)) + 1.0f);
```

that generates each data point, that's the best I can do without doing, say, nonlinear least squares in the test itself to find it.

aliddell: Honestly, the test itself doesn't demonstrate anything. Given the [expression](https://github.com/acquire-project/acquire-driver-common/blob/main/src/simcams/imfill.pattern.cpp#L54) that generates the radial sine pattern

```cpp
(T)(127.0f * (sinf(6.28f * (t * 10.0f + (dx2 + dy2) * 1e-2f)) + 1.0f));
```

I had been hoping to find `t` numerically from the data but without a good starting point my methods are failing me. I move we just get rid of the test.

examples/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
+if (${WITH_EXAMPLES})
```
nclack, line 1: Also see [option](https://cmake.org/cmake/help/latest/command/option.html) for a way to get this flag into the config. It gives users a checkbox in cmake-gui and a way to provide a tooltip.

PR #95, opened 2023-08-15
Don't build tests of tests

aliddell:
- Updates acquire-driver-common to latest
- Set `NOTEST` for test submodules


nclack:
looks like the action needs to be changed to not do recursive submodule inits

PR #101, opened 2023-08-17
Implement and test basic Zarr v3

aliddell:
Data generated in `write-zarr-v3-raw.cpp` and `write-zarr-v3-compressed.cpp` were tested as readable using zarr-python.


src/zarr.cpp
```diff
@@ -770,6 +778,133 @@ zarr::Zarr::recover_threads_()
...
+                { "clevel", params.clevel_ },
+                { "cname", params.codec_id_ },
+                { "shuffle", params.shuffle_ } } },
+        };
+    }
+
+    auto path = (fs::path(data_root_) / "meta" / "root" /
+                 (std::to_string(level) + ".array.json"))
+                  .string();
+    write_string(path, metadata.dump(4));
```
andy-sweet, line 850: Optional: not really part of this PR, but I think we should consider passing through the `fs::path` object instead of converting to a `string` here and elsewhere because the first thing we do in `write_string` is convert it back to a path.

src/zarr.cpp
```diff
@@ -770,6 +778,133 @@ zarr::Zarr::recover_threads_()
...
+/// @details Zarr v3 stores group metadata in
+/// /meta/{group_name}.group.json. We will call the group "root".
+void
+zarr::ZarrV3::write_group_metadata_() const
+{
+    namespace fs = std::filesystem;
+    using json = nlohmann::json;
+
+    json metadata;
+    metadata["attributes"]["acquire"] = json::parse(external_metadata_json_);
```
andy-sweet, line 890: Is it part of the Zarr v3 standard that it's expected to store extra/external metadata in the group attributes (i.e. anyone can put anything they want here)? Or are we assuming this should be a safe enough namespace?

src/zarr.hh
```diff
@@ -132,6 +134,27 @@ struct Zarr final : StorageInterface
     void recover_threads_();
 };
 
+struct ZarrV3 final : public Zarr
```
andy-sweet, line 137: Optional: don't feel strongly enough to really suggest this, but I wonder if we should really use inheritance here. Seems like it's mostly being used to share code rather storing heterogenous collections of zarr and zarrv3 objects and I generally feel like plain old functions are typically better for sharing code.

andy-sweet (commented):
Took a first pass through and I don't see any big issues. Will take another pass before approving/requesting changes.

src/zarr.cpp
aliddell: Yeah, that's my understanding from [this](https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#attributes).

andy-sweet (approved):
After a second pass, looks good to me. Any remaining comments are optional suggestions, some of which may be better handled outside of this PR.

src/zarr.cpp
aliddell: No, I think this is a fine idea.

PR #113, opened 2023-08-25
Fix runaway job queue

### LINKED ISSUE #112: Zarr hangs for longer than 30 seconds at the end of large acquisitions when multiscale is enabled

aliddell:
Closes #112 and #114.

### DIFF
--- a/src/chunk.writer.cpp
+++ b/src/chunk.writer.cpp
@@ -111,6 +111,9 @@ ChunkWriter::ChunkWriter(BaseEncoder* encoder,
 ChunkWriter::~ChunkWriter()
 {
     close_current_file();
+    std::stringstream ss;
+    if (!std::is_sorted(frame_ids_.begin(), frame_ids_.end()))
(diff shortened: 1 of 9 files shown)


src/chunk.writer.cpp
```diff
@@ -190,14 +191,47 @@ ChunkWriter::write(const uint8_t* beg, const uint8_t* end)
...
+
+        EXPECT(nbytes == bpt,
+               "Expected to write %lu bytes, but wrote %lu bytes.",
+               bpt,
+               nbytes);
+
+        last_frame_ = frame->frame_id();
+    }
+
+    frames_.clear();
```
nclack, line 224: this is a problem. Need to remove the frames that got processed and leave the rest. Might be easiest to switch to a priority queue.

nclack (commented):
There's the issue with the frames_ clear in `write_frames` that needs a fix. Also, left some nits.

PR #117, opened 2023-08-29
Catch exceptions off the main thread

aliddell:
- Makes `ChunkWriter::write_frame()` and `FrameScaler::push_frame()` both `noexcept`.
- Changes an in-thread call to `filesystem::create_directories` to use the non-throwing version.

### DIFF
--- a/src/chunk.writer.cpp
+++ b/src/chunk.writer.cpp
@@ -166,8 +166,9 @@ ChunkWriter::write(const uint8_t* beg, const uint8_t* end)
     if (0 == bytes_in)
         return 0;
 
-    if (!current_file_.has_value())
-        open_chunk_file();
+    if (!current_file_.has_value()) {
+        CHECK(open_chunk_file());
+    }
 
     size_t bytes_out = 0;
     auto* cur = (uint8_t*)beg;
--- a/src/chunk.writer.cpp
+++ b/src/chunk.writer.cpp
@@ -115,20 +115,28 @@ ChunkWriter::~ChunkWriter()
 }
 
 bool
-ChunkWriter::write_frame(const TiledFrame& frame)
+ChunkWriter::write_frame(const TiledFrame& frame) noexcept
 {
     std::scoped_lock lock(mutex_);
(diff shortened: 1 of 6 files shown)


src/chunk.writer.cpp
```diff
@@ -210,13 +219,26 @@ ChunkWriter::open_chunk_file()
...
-        fs::create_directories(parent_path);
+    if (!fs::is_directory(parent_path)) {
+        std::error_code ec;
+        // the call may fail if the directory is being created in another thread
+        if (!fs::create_directories(parent_path, ec) &&
+            !fs::is_directory(parent_path)) {
+            LOGE("Failed to create directory '%s': %s",
+                 parent_path.string().c_str(),
+                 ec.message().c_str());
+            return false;
```
nclack, line 230: I like `create_directories`. 

This looks like it fails if another thread is creating the folder at the same time. What happens then? How does it recover?

andy-sweet: If I understood correctly, then we'll skip this chunk in the failing case. I don't know if/how zarr readers deal with missing chunks, but I guess we don't want to have missing chunks if we can avoid it. I guess we could retry, but I guess that will complicate the scheduling/ordering logic.

Another thought is to guard calls to `create_directories` with a lock, but that obviously comes with a performance hit that I can't really analyze without more data.

Or could we create all the directories we need on the main scheduling thread before we start writing the chunks?

andy-sweet (approved):
Code changes look good. I made some alternate suggestions, but I don't think they block this PR. I can't test this today with the reliable reproducer I had, but can do tomorrow.

src/chunk.writer.cpp
nclack: > failure in a thread should be a fatal error

Yes. Any failure to record data should stop the acquisition.

nclack: There are a couple approaches I can imagine here. 

One is to retry for a while until success. The idea is that this block has to ensure a writable directory for the next part of the code. There should be an associated timeout.

Two is to create a unique temporary directory and then move it later? That seems like it can't work but I thought I'd throw it out there.

PR #125, opened 2023-09-21
76 sharding in zarr

### LINKED ISSUE #76: Implement sharding in Zarr
## Description

[Spec](https://zarr-specs.readthedocs.io/en/latest/v3/codecs/sharding-indexed/v1.0.html)

## Definition of done

- We are able to read a sharded array (written by Acquire) in zarr-python.

### LINKED ISSUE #111: Basic Zarr v3

### LINKED ISSUE #112: Zarr hangs for longer than 30 seconds at the end of large acquisitions when multiscale is enabled

### LINKED ISSUE #114: Zarr occasionally writes frames out of order
It's rare, but it happens due to the way the frames are dispatched to the chunk writers.

### LINKED ISSUE #115: Zarr can raise exceptions off the main thread when creating directories
Observed when writing a [spinnaker (blackfly) + zarr integration test on Windows](https://gist.github.com/andy-sweet/12653977cc51643b8a2b90c6cedd9591). This was flaky, but I got pretty predictable failures with 100 frames.

The test aborted when the main thread was calling `acquire_stop`. The zarr writer is in the process of emptying its queue of tasks, but has a while to go (the queue has a size of 36). See the [output logs for more details about progress](https://gist.github.com/andy-sweet/d373be86be2d2698734fa8da1bdc81ed).

One of the zarr writer threads threw using filesystem_error. It was attempting to create `out.zarr\\0/8/0/0` (when only `out.zarr/0|1|2|3` were created so far).  When calling `filesystem::create_directories` in the Windows std implementation, we got `error_code._MyVal == 3`.

Not sure why the error is occurring or how to fix it, but we should ensure that we're handling exceptions on the zarr worker threads so that they can be reported as error codes.

aliddell:
Implements Zarr v3 with the [sharding storage transformer](https://web.archive.org/web/20230213221154/https://zarr-specs.readthedocs.io/en/latest/extensions/storage-transformers/sharding/v1.0.html). (Not the [sharding codec](https://zarr-specs.readthedocs.io/en/latest/v3/codecs/sharding-indexed/v1.0.html), which is WIP, but I'm keeping an eye on it.)

Supersedes #101. Supersedes #113. Supersedes #117.
Closes #76. Closes #111. Closes #112. Closes #114. Closes #115.

### Added

- Support for [Zarr v3](https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html).
- Support for
  the [sharding storage transformer](https://web.archive.org/web/20230213221154/https://zarr-specs.readthedocs.io/en/latest/extensions/storage-transformers/sharding/v1.0.html)
  in Zarr v3.
- Ship debug libs for C-Blosc on Linux and Mac.

### Changed

- Upgrades C-Blosc from v1.21.4 to v1.21.5.

### Fixed

- A bug where enabling multiscale without specifying the tile size would cause an error.
- Exceptions thrown off the main thread are now caught and logged, and Zarr throws an error in `append`.
- Job queue is now cleared after every operation.


src/common.cpp
```diff
@@ -0,0 +1,95 @@
...
+{
+    return bytes_per_tile(tile_shape, type) *
+           frames_per_chunk(tile_shape, type, max_bytes_per_chunk);
+}
+
+const char*
+common::sample_type_to_dtype(SampleType t)
+
+{
+    static const char* table[] = { "u1", "u2", "i1", "i2",
```
andy-sweet, line 58: Optional: slightly risky to look up the sampletype -> dtype correspondence by index rather than having an explicit map or switch/case and performance doesn't seem critical here but I don't feel too strongly.

src/common.cpp
```diff
@@ -0,0 +1,95 @@
...
+                                   "f4", "u2", "u2", "u2" };
+    if (t < countof(table)) {
+        return table[t];
+    } else {
+        throw std::runtime_error("Invalid sample type.");
+    }
+}
+
+const char*
+common::sample_type_to_string(SampleType t) noexcept
```
andy-sweet, line 68: Nit: should probably be defined in `acquire-core-libs`, but no problem being here for now.

src/writers/blosc.writer.cpp
```diff
@@ -0,0 +1,3 @@
+//
+// Created by Alan Liddell on 2023-09-18.
```
andy-sweet, line 2: Why is this otherwise empty?

src/writers/blosc.writer.hh
```diff
@@ -0,0 +1,4 @@
+#ifndef ACQUIRE_DRIVER_ZARR_BLOSC_WRITER_HH
+#define ACQUIRE_DRIVER_ZARR_BLOSC_WRITER_HH
+
```
andy-sweet, line 3: Why is this otherwise empty?

src/common.hh
```diff
@@ -0,0 +1,62 @@
...
+
+namespace common {
+size_t
+bytes_of_type(const SampleType& type);
+
+size_t
+bytes_per_tile(const ImageDims& tile_shape, const SampleType& type);
+
+size_t
+frames_per_chunk(const ImageDims& tile_shape,
```
andy-sweet, line 32: Optional: I know this name isn't new, but I wonder if `tiles_per_chunk` would make these things and their calculations a little easier to reason about.

andy-sweet (commented):
I'm still making my way through this, but haven't seen anything blocking so far. About to travel for ~45 minutes where I can probably review this with Nathan.

src/writers/blosc.writer.hh
aliddell: Good catch, must've been from a `git add .`.

src/common.cpp
aliddell: [I agree.](https://github.com/acquire-project/acquire-core-libs/pull/36)

tests/write-zarr-with-defaults.cpp
```diff
@@ -134,9 +134,9 @@ acquire(AcquireRuntime* runtime, const char* filename)
         do {
             struct clock throttle;
             clock_init(&throttle);
-            EXPECT(clock_cmp_now(&clock) < 0,
-                   "Timeout at %f ms",
-                   clock_toc_ms(&clock) + time_limit_ms);
+//            EXPECT(clock_cmp_now(&clock) < 0,
```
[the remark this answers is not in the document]
aliddell: Thanks for catching this.

src/common.hh
aliddell: It's a tossup for me. If Nathan seconds this, I'll change it.

PR #135, opened 2023-10-12
Zarr many dims

andy-sweet:
This is a very rough effort to get modify Acquire's zarr writer to output many dimensions.

It requires the user to specify exactly how many channels and slices there are in a z-stack. Chunks only roll over in the z-dimension.


andy-sweet:
This has outlived its usefulness, so closing.

PR #136, opened 2023-10-19
Breakout Zarr v2 files from #76

### LINKED ISSUE #112: Zarr hangs for longer than 30 seconds at the end of large acquisitions when multiscale is enabled

### LINKED ISSUE #114: Zarr occasionally writes frames out of order

### LINKED ISSUE #115: Zarr can raise exceptions off the main thread when creating directories

aliddell:
First half of #125.

Supersedes #113. Supersedes #117.
Closes #112. Closes #114. Closes #115.


src/writers/writer.hh
```diff
@@ -0,0 +1,117 @@
...
+    ImageDims frame_dims_;
+    ImageDims tile_dims_;
+    uint16_t tiles_per_frame_x_;
+    uint16_t tiles_per_frame_y_;
+    SampleType pixel_type_;
+    uint32_t frames_per_chunk_;
+
+    /// Compression
+    std::optional<BloscCompressionParams> blosc_compression_params_;
+    // std::optional<ZstdCompressionParams> zstd_compression_params_; // TODO
```
andy-sweet, line 80: Is this TODO inherited by a code move? Can we just remove it?

src/writers/writer.hh
```diff
@@ -0,0 +1,117 @@
...
+
+    /// Filesystem
+    FileCreator file_creator_;
+    fs::path data_root_;
+    std::vector<file> files_;
+
+    /// Multithreading
+    std::vector<std::vector<uint8_t>> chunk_buffers_;
+    bool* buffers_ready_;
+    std::mutex mutex_;
```
andy-sweet, line 90: What is the mutex being used to guard? I think it's mostly for `chunk_buffers_`, but maybe `buffers_ready_` too?

Could that be communicated in its name? E.g. `buffers_mutex_`.

I often find being specific and explicit about what is intended to be guarded pretty useful, even if it's just in the form of a comment. I really like static analysis for things like this and things like Java's [GuardedBy](https://errorprone.info/bugpattern/GuardedBy), but I'm unaware of something we can use for C++ and that's definitely overkill for this PR.

src/zarr.v2.cpp
```diff
@@ -0,0 +1,268 @@
...
+              { "coordinateTransformations",
+                {
+                  {
+                    { "type", "scale" },
+                    {
+                      "scale",
+                      {
+                        std::pow(2, i), // t
+                                        // TODO (aliddell): c?
+                        1,              // z
```
andy-sweet, line 212: We don't include `z` right now, which is [optional in the multiscales ome-ngff spec](https://ngff.openmicroscopy.org/latest/#multiscale-md)). So the 1 actually corresponds to the channel axis (to accommodate multi-channel cameras).

I guess this was probably inherited from moving code, but seems like a good time to delete it.

src/zarr.v2.cpp
```diff
@@ -0,0 +1,268 @@
...
+            image_dims.rows, // y
+            image_dims.cols, // x
+          } },
+        { "chunks",
+          {
+            frames_per_chunk, // t
+                              // TODO (aliddell): c?
+            1,                // z
+            tile_dims.rows,   // y
+            tile_dims.cols,   // x
```
andy-sweet, line 114: See related comment below. Swap z for c.

src/zarr.hh
```diff
@@ -1,41 +1,27 @@
 #ifndef H_ACQUIRE_STORAGE_ZARR_V0
 #define H_ACQUIRE_STORAGE_ZARR_V0
 
+#ifndef __cplusplus
+#error "This header requires C++20"
+#endif
+
```
andy-sweet, line 7: Optional

Given that the implementation is quite a bit more complex now (so that we can easily accommodate v2 and v3 writers), I'd really value a short description of the key components of the implementation and how they're related - really just a list of class names and one line summaries.

This seems like a good place for this (i.e. it's one of the first places I would look) or alternatively a `README` in `src.

andy-sweet (approved):
Looks good to me. Most of my comments are minor additions or optional.

We've looked at this or similar a few times and I think there's enough understanding to get it in and move forward. There are still some parts of the code and execution order I don't have a good grasp on, but that can be improved in the future (both by me and by iterating on the code itself).

src/zarr.hh
aliddell: I added a README here.

PR #137, opened 2023-10-19
Zarrv3 part two

### LINKED ISSUE #76: Implement sharding in Zarr

### LINKED ISSUE #111: Basic Zarr v3

aliddell:
Implements Zarr v3 with the [sharding storage transformer](https://web.archive.org/web/20230213221154/https://zarr-specs.readthedocs.io/en/latest/extensions/storage-transformers/sharding/v1.0.html). (Not the [sharding codec](https://zarr-specs.readthedocs.io/en/latest/v3/codecs/sharding-indexed/v1.0.html), which is WIP, but I'm keeping an eye on it.)

Supersedes #101. Supersedes #125.
Closes #76. Closes #111.

### Added

- Support for [Zarr v3](https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html).
- Support for
  the [sharding storage transformer](https://web.archive.org/web/20230213221154/https://zarr-specs.readthedocs.io/en/latest/extensions/storage-transformers/sharding/v1.0.html)
  in Zarr v3.


aganders3:
For what it's worth I'm able to compile, tests pass, and I can read in the test data using zarr-python w/ experimental V3 and sharding support. Promising so far!

README.md
```diff
@@ -24,6 +27,10 @@ Chunking is configured using `storage_properties_set_chunking_props()` when conf
 Multiscale storage can be enabled or disabled by calling `storage_properties_set_enable_multiscale()` when configuring
 the video stream.
 
+For the [Zarr v3] version of each device, you can use the `ZarrV3*` devices.
```
aliddell, line 30: Add a note here about experimental flags in zarr-python.

src/zarr.v3.cpp
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    const auto w_rat = (float)frame_dims.cols / (float)tile_dims.cols;
+    auto shard_cols = (uint32_t)std::ceil(w_rat * tile_dims.cols);
+    if (shard_cols > frame_dims.cols) {
+        auto n_shards_cols = smallest_prime_factor(shard_cols / tile_dims.cols);
+        shard_dims.cols = n_shards_cols * tile_dims.cols;
+    }
+
+    return shard_dims;
```
aliddell, line 83: Document this.

src/writers/shard.writer.cpp
```diff
@@ -0,0 +1,292 @@
...
+            const auto k = i * chunks_per_shard + j;
+            const auto& chunk = chunk_buffers_.at(k);
+            memcpy(shard.data() + offset, chunk.data(), chunk_sizes.at(k));
+            offset += chunk_sizes.at(k);
+        }
+        memcpy(shard.data() + offset,
+               chunk_indices.data(),
+               chunk_indices.size() * 8);
+        offset += chunk_indices.size() * 8;
+        shard_sizes.push_back(offset);
```
aliddell, line 222: Document the steps here.

src/writers/shard.writer.cpp
```diff
@@ -0,0 +1,292 @@
...
+            shard.resize(shard_size + index_size);
+        }
+
+        size_t offset = 0;
+        for (auto j = 0; j < chunks_per_shard; ++j) {
+            const auto k = i * chunks_per_shard + j;
+            const auto& chunk = chunk_buffers_.at(k);
+            memcpy(shard.data() + offset, chunk.data(), chunk_sizes.at(k));
+            offset += chunk_sizes.at(k);
+        }
```
aliddell, line 217: Just shard buffers?

src/writers/shard.writer.cpp
```diff
@@ -0,0 +1,292 @@
...
+            chunk_indices.push_back(shard_size); // chunk index
+            const auto k = i * chunks_per_shard + j;
+            shard_size += chunk_sizes.at(k);
+            chunk_indices.push_back(chunk_sizes.at(k)); // chunk extent
+        }
+
+        // if we're very unlucky we can technically run into this
+        if (shard.size() < shard_size + index_size) {
+            shard.resize(shard_size + index_size);
+        }
```
aliddell, line 209: Avoid this by allocating nchunks * (bytes_per_chunk + BLOSC_MAX_OVERHEAD)

src/writers/shard.writer.cpp
```diff
@@ -0,0 +1,292 @@
...
+            offset += chunk_sizes.at(k);
+        }
+        memcpy(shard.data() + offset,
+               chunk_indices.data(),
+               chunk_indices.size() * 8);
+        offset += chunk_indices.size() * 8;
+        shard_sizes.push_back(offset);
+    }
+
+    // write out
```
aliddell, line 225: Use a latch?

src/writers/shard.writer.cpp
```diff
@@ -0,0 +1,292 @@
...
+        }
+
+        std::fill(buf.begin(), buf.end(), 0);
+    }
+    const auto bytes_per_shard = bytes_per_chunk * chunks_per_shard;
+    for (auto& buf : shard_buffers_) {
+        // absurd edge case we need to account for
+        if (buf.size() > bytes_per_shard + index_size) {
+            buf.resize(bytes_per_shard + index_size);
+        }
```
aliddell, line 279: Avoid resize?

src/zarr.v3.cpp
```diff
@@ -0,0 +1,299 @@
...
+    } else {
+        writers_.push_back(std::make_shared<ShardWriter>(
+          frame_dims,
+          shard_dims_,
+          tile_dims,
+          (uint32_t)(max_bytes_per_chunk_ / bytes_per_tile),
+          (get_data_directory_() / "0").string(),
+          this));
+    }
+}
```
aliddell, line 121: LOD

src/zarr.v3.cpp
```diff
@@ -0,0 +1,299 @@
...
+      { "type", "regular" },
+    });
+    metadata["chunk_memory_layout"] = "C";
+    metadata["data_type"] = common::sample_type_to_dtype(pixel_type_);
+    metadata["extensions"] = json::array();
+    metadata["fill_value"] = 0;
+    metadata["shape"] = json::array({
+      frame_count,     // t
+                       // TODO (aliddell): c?
+      1,               // z
```
aliddell, line 182: fix

aliddell:
The tale grew in the telling.

aganders3:
These updates look good to me. Thanks!

I was surprised to see performance much worse on the tests, but it was only for the compressed. For anyone else who notices I think it's from the change of test pattern, not the code.

src/writers/writer.cpp
```diff
@@ -145,21 +118,65 @@ zarr::FileCreator::create_channel_dirs_(int n_c) noexcept
                            "Failed to create directory: %s",
                            exc.what());
                   err = buf;
+                  failure = true;
```
nclack, line 121: If there is one failure, should we short-circuit other waiting jobs? I imagine failures are often correlated, and we might not want to wait for all of them to fail. Could just check to see if failure is set right before the expensive part, for example.

nclack (commented):
Added some comments.

src/common.hh
```diff
@@ -42,19 +43,38 @@ struct ImageDims
...
-        };
-};
+struct Zarr;
 
 namespace common {
+struct ThreadPool final
+{
+  public:
+    using JobT = std::function<bool(std::string&)>;
+    struct ThreadContext
```
nclack, line 53: this struct can be private, I think?

src/common.cpp
```diff
@@ -1,10 +1,93 @@
...
+common::ThreadPool::pop_from_job_queue_() noexcept
+{
+    std::scoped_lock lock(jobs_mutex_);
+    if (jobs_.empty()) {
+        return std::nullopt;
+    }
+
+    auto job = jobs_.front();
+    jobs_.pop();
+    return job;
```
nclack, line 52: Is `front()` still live after `pop()`? May need to use `swap()` to swap in a dummy value - or copy the value out before pop.

nclack: In `thread_worker_`, I want to assume that the returned job is independent of the queue once returned from this function.

src/common.cpp
```diff
@@ -1,10 +1,93 @@
...
+        }
+
+        // FIXME: wait
+        if (auto job = pop_from_job_queue_(); job.has_value()) {
+            ctx->ready = false;
+            if (std::string err_msg; !job.value()(err_msg)) {
+                error_handler_(err_msg);
+            }
+            ctx->ready = true;
+            lock.unlock();
```
nclack, line 81: It looks like this lock is just protecting the job queue. If that's the case, it doesn't need to be held while the job is executing. This would allow jobs to be added to the queue even when worker threads are busy.

src/common.cpp
```diff
@@ -1,10 +1,93 @@
...
+        if (ctx->should_stop) {
+            break;
+        }
+
+        // FIXME: wait
+        if (auto job = pop_from_job_queue_(); job.has_value()) {
+            ctx->ready = false;
+            if (std::string err_msg; !job.value()(err_msg)) {
+                error_handler_(err_msg);
+            }
```
nclack, line 79: Just pointing out the non-standard if statement here. I can't tell whether I like yours or the suggestion more.

Maybe only invoke the `error_handler_` on a non-empty `err_msg`?

```suggestion
            {
               std::string err_msg;
               if (!job.value()(err_msg)) {
                   error_handler_(err_msg);
               }
            }
```

src/common.hh
```diff
@@ -21,7 +43,38 @@ struct ImageDims
...
+    {
+        std::thread thread;
+        std::mutex mutex;
+        std::condition_variable cv;
+        bool should_stop;
+        bool ready;
+    };
+
+    ThreadPool(size_t n_threads, std::function<void(const std::string&)> err);
```
nclack, line 62: Need to document the error handler by adding a comment here. When's it called and what is the recommended behavior? Any constraints?

src/common.cpp
```diff
@@ -1,10 +1,93 @@
...
+}
+
+common::ThreadPool::~ThreadPool() noexcept
+{
+    // spin down threads
+    for (auto& ctx : contexts_) {
+        ctx.should_stop = true;
+        ctx.cv.notify_one();
+        ctx.thread.join();
+    }
```
nclack, line 32: Very very often, I find it necessary to pull blocking behavior like this outside of the destructor - into a `await_stop()` kind of function. Then call that at the appropriate point in the client code.

The destructor's behavior should be more severe - aborting threads without waiting for them to complete if possible.

The reason is that we want to be able to handle exceptions from the `stop()` in a controlled way, and that's hard at destruct-time.

src/common.cpp
```diff
@@ -1,10 +1,93 @@
...
+
+    while (true) {
+        std::unique_lock lock(ctx->mutex);
+        ctx->cv.wait_for(lock, 1ms, [&] { return ctx->should_stop; });
+
+        if (ctx->should_stop) {
+            break;
+        }
+
+        // FIXME: wait
```
nclack, line 74: Need to address the FIXME.

I think this is here because this loop will excessively spin. I have some examples of how I write polling loops in runtime - you can search for `throttler_wait`.

src/common.cpp
```diff
@@ -1,10 +1,93 @@
...
+        if (auto job = pop_from_job_queue_(); job.has_value()) {
+            ctx->ready = false;
+            if (std::string err_msg; !job.value()(err_msg)) {
+                error_handler_(err_msg);
+            }
+            ctx->ready = true;
+            lock.unlock();
+            ctx->cv.notify_one();
```
nclack, line 82: The `cv` should get notified when the code that just run might have make the associated predicate false. I want the predicate to be `!is_stopping && jobs.empty()`. In this case, we haven't changed the stopping state or added anything to the queue, so there's no need to notify.

src/writers/shard.writer.cpp
```diff
@@ -0,0 +1,158 @@
+#include "shard.writer.hh"
+#include "../zarr.hh"
+
+#include <latch>
+#include <stdexcept>
+
+namespace zarr = acquire::sink::zarr;
+
+zarr::ShardWriter::ShardWriter(const ImageDims& frame_dims,
```
nclack, line 9: Can we rename `ShardWriter` and `ChunkWriter` in this PR?

src/writers/writer.hh
```diff
@@ -26,20 +22,21 @@ struct Zarr;
 struct FileCreator
 {
     FileCreator() = delete;
-    explicit FileCreator(Zarr* zarr);
+    explicit FileCreator(std::shared_ptr<common::ThreadPool> thread_pool);
     ~FileCreator() noexcept = default;
 
-    void set_base_dir(const fs::path& base_dir) noexcept;
-    [[nodiscard]] bool create(int n_c,
+    [[nodiscard]] bool create(const fs::path& base_dir,
```
nclack, line 28: I'd prefer a more descriptive name.
```suggestion
    [[nodiscard]] bool create_dirs(const fs::path& base_dir,
```

src/writers/writer.hh
```diff
@@ -84,30 +82,27 @@ struct Writer
...
-                                size_t buf_size) noexcept = 0;
+    void compress_buffers_() noexcept;
+    size_t write_frame_to_chunks_(const uint8_t* buf, size_t buf_size) noexcept;
     virtual void flush_() noexcept = 0;
 
     uint32_t tiles_per_frame_() const;
 
     /// Files
-    [[nodiscard]] virtual bool make_files_() noexcept = 0;
+    [[nodiscard]] virtual bool make_files_() noexcept = 0; // FIXME: pull down
```
nclack, line 105: stray FIXME

nclack (commented):
I left some comments on the `ThreadPool`. There are also some stray FIXME's.

src/common.cpp
aliddell: It is still live, but not for any good reason. `jobs_.front()` returns a reference, true, but we're invoking the copy constructor here. If it will make you less uneasy, I can `move` it.

src/common.hh
aliddell: Yeah, so it can.

src/common.cpp
aliddell: Yeah, this is vestigial. Good catch.

src/common.cpp
aliddell: Functionally they're the same.

Why would we only invoke the error handler when `err_msg` is nonempty? The job returns a bool, and if it returns `false` the Zarr writer should fail on the next `append` call. The error message is just diagnostic.

src/common.cpp
aliddell: Yeah, more vestigial code. There's no need for `ready` at all, in fact.

src/common.cpp
aliddell: I'm not sure how to abort threads without just calling `thread.join()` on them, which will block. Looking around I see Windows-specific solutions. Still, good point about stopping the threads before destructing. I'll take care of that.

src/common.cpp
```diff
@@ -1,10 +1,95 @@
...
+    while (true) {
+        std::unique_lock lock(ctx->mutex);
+        ctx->cv.wait(lock, [&] { return ctx->should_stop || !jobs_.empty(); });
+
+        if (ctx->should_stop) {
+            break;
+        }
+
+        if (auto job = pop_from_job_queue_(); job.has_value()) {
+            if (std::string err_msg; !job.value()(err_msg)) {
```
nclack, line 85: unlock here

aliddell:
Don't merge yet. One of the tests is segfaulting on Mac, but only sometimes.

src/common.cpp
```diff
@@ -1,10 +1,93 @@
...
+    await_stop();
+}
+
+void
+common::ThreadPool::push_to_job_queue(JobT&& job)
+{
+    std::scoped_lock lock(*jobs_mutex_);
+    jobs_.push(std::move(job));
+
+    for (auto& ctx : contexts_) {
```
andy-sweet, line 38: Optional: this `for` loop pattern with `notify_one` seems a bit weird to me, as we know one job has been added here, but we can potentially notify all the threads in the pool. I think a viable alternative solution is only have one condition variable that's associated with the `ThreadPool`, wait on that in `thread_worker_` which should have a live `ThreadPool` instance, then just call `notify_one` here so at most one thread is notified.

This is optional because notifying all of them works, but only the first waiting one will win, so it's sub-optimal.

src/common.cpp
```diff
@@ -1,10 +1,93 @@
...
+
+void
+common::ThreadPool::await_stop() noexcept
+{
+    // spin down threads
+    for (auto& ctx : contexts_) {
+        ctx.should_stop = true;
+        ctx.cv.notify_one();
+
+        if (ctx.thread.joinable()) {
```
andy-sweet, line 51: Would we ever expect a non-joinable thread in this thread pool?

src/common.cpp
```diff
@@ -1,10 +1,93 @@
...
+    }
+}
+
+void
+common::ThreadPool::await_stop() noexcept
+{
+    // spin down threads
+    for (auto& ctx : contexts_) {
+        ctx.should_stop = true;
+        ctx.cv.notify_one();
```
andy-sweet, line 49: Optional: if we only need one condition variable (as suggested in another comment), then I think we may only need one `ThreadPool.should_stop` in which case `notify_all` could be used instead here.

This is optional for the same reason the other suggestion is - what you have works, but this may be simpler.

andy-sweet (approved):
As suggested by @nclack, I only reviewed the `ThreadPool` class. I made a few optional potentially simplifying suggestions but none of them are blocking given what you have works and I'm not certain I'm thinking through all intended usage or the lifetime of `ThreadPool`. Happy to chat over video/audio to clarify those. As `ThreadPool` is nicely separated we could always simplify it later if desired.

src/common.cpp
aliddell: Yes, because `await_stop` is called from `Zarr::stop`, which is called at least once during setup before acquisition actually happens.

src/common.cpp
aliddell: No, this makes a lot more sense than what I have and is much simpler.

src/common.cpp
```diff
@@ -1,10 +1,93 @@
...
+  , error_handler_{ err }
+{
+    if (n_threads == 0) {
+        throw std::runtime_error("Cannot create thread pool with 0 threads.");
+    }
+
+    for (auto& ctx : contexts_) {
+        ctx.should_stop = false;
+        ctx.thread = std::thread([this, ctx_ = &ctx] { thread_worker_(ctx_); });
+        ctx.mutex = jobs_mutex_;
```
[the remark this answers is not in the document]
aliddell: That is a fair point.

src/common.cpp
```diff
@@ -1,10 +1,93 @@
...
+
+common::ThreadPool::~ThreadPool() noexcept
+{
+    await_stop();
+}
+
+void
+common::ThreadPool::push_to_job_queue(JobT&& job)
+{
+    std::scoped_lock lock(*jobs_mutex_);
```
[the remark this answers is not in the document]
aliddell: No, I think this is fair. It may actually be responsible for the sporadic failures I've been seeing in the tests.

PR #158, opened 2023-11-16
fix test timeout

aliddell:
A glance at the [actions](https://github.com/acquire-project/acquire-driver-zarr/actions) lately shows that all testing is timing out on Windows only, and on `test-get-meta` only. I can't reproduce this on my machine, but I suspect this is because threads are being shut down before they have a chance to properly start, since we're not actually doing anything with them in that test, so they're getting left in a bad state where they're joinable but we can't join them. So instead of spinning up threads on the thread pool's construction, I do it in `Zarr::start`. That seems to fix the problem.

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -11,18 +11,21 @@ common::ThreadPool::ThreadPool(size_t n_threads,
                                std::function<void(const std::string&)> err)
   : error_handler_{ err }
 {
-    if (n_threads == 0) {
-        throw std::runtime_error("Cannot create thread pool with 0 threads.");
-    }
-
-    for (auto i = 0; i < n_threads; ++i) {
-        threads_.emplace_back([this] { thread_worker_(); });
-    }
+    n_threads_ = std::clamp(
+      n_threads, (size_t)1, (size_t)std::thread::hardware_concurrency());
 }
 
 common::ThreadPool::~ThreadPool() noexcept
 {
     await_stop();
+}
+
+void
+common::ThreadPool::start()
+{
+    for (auto i = 0; i < n_threads_; ++i) {
+        threads_.emplace_back([this] { thread_worker_(); });
+    }
 }
 
 void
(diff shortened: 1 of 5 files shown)


nclack:
Should there be something on the `stop()` side? What happens if `start()` gets called multiple times?

src/common.cpp
```diff
@@ -11,20 +11,23 @@ common::ThreadPool::ThreadPool(size_t n_threads,
...
 {
-    if (n_threads == 0) {
-        throw std::runtime_error("Cannot create thread pool with 0 threads.");
-    }
-
-    for (auto i = 0; i < n_threads; ++i) {
-        threads_.emplace_back([this] { thread_worker_(); });
-    }
+    n_threads_ = std::clamp(
+      n_threads, (size_t)1, (size_t)std::thread::hardware_concurrency());
```
andy-sweet, line 15: It seems possible that [`hardware_concurrency` can return 0](https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency) and the [behavior of `clamp` is undefined when `lo > hi`](https://en.cppreference.com/w/cpp/algorithm/clamp).

So consider guarding against that (likely rare) case, by checking the return value of `hardware_concurrency` and/or using `max` of it and 1.

src/common.cpp
```diff
@@ -11,20 +11,23 @@ common::ThreadPool::ThreadPool(size_t n_threads,
...
+common::ThreadPool::start()
+{
+    for (auto i = 0; i < n_threads_; ++i) {
+        threads_.emplace_back([this] { thread_worker_(); });
+    }
+}
+
 void
 common::ThreadPool::push_to_job_queue(JobT&& job)
 {
```
andy-sweet, line 33: Optional: might want to consider checking if this has been started and similarly if it has been stopped, likely raising in both cases.

andy-sweet (changes requested):
I'm not used to explicitly starting thread pools, but I don't hate it and it's not used much, so this seems fine to unblock the upcoming release.

My only request is around a little more guarding around the `clamp` operation in the `ThreadPool` constructor.

andy-sweet (approved):
Looks good now!

PR #161, opened 2023-11-17
Fix test timeout redux

aliddell:
- Fixes a bug in the Zarr V3 writer where external metadata was tacitly assumed to be nonempty.
- Removes explicit thread start from `ThreadPool` in favor of start on construction.
- Aborts thread jobs in the destructor only, allowing any job that's made it on to the queue to finish before shutting down the thread pool in the usual case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.1.5](https://github.com/acquire-project/acquire-driver-zarr/compare/v0.1.3...v0.1.4) - 2023-11-17
+## [0.1.5](https://github.com/acquire-project/acquire-driver-zarr/compare/v0.1.3...v0.1.4) - 2023-11-20
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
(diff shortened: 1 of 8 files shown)


src/common.hh
```diff
@@ -60,24 +60,22 @@ struct ThreadPool final
...
     std::vector<std::thread> threads_;
     mutable std::mutex jobs_mutex_;
     std::condition_variable cv_;
     std::queue<JobT> jobs_;
 
-    bool started_;
-    std::atomic<bool> should_stop_;
+    std::atomic<bool> is_accepting_jobs_;
 
     /// Multithreading
```
andy-sweet, line 76: Optional

This comment is not very informative. I think it's telling me that these methods are expected be executed on the worker threads this pool owns (and not on the client/calling thread). The other useful information is that the caller of these methods is expected to lock the require mutex to guarantee thread safety and consistency. So maybe update this along those lines?

src/common.hh
```diff
@@ -60,24 +60,22 @@ struct ThreadPool final
     ThreadPool(size_t n_threads, std::function<void(const std::string&)> err);
     ~ThreadPool() noexcept;
 
-    void start();
     void push_to_job_queue(JobT&& job);
     void await_stop() noexcept;
```
andy-sweet, line 64: Optional

Add a comment here to explain that calling this will wait/block until pending jobs are complete. Also, after this is complete, no new jobs can be submitted.

andy-sweet (approved):
Looks good to me now.

src/common.hh
aliddell: I'll just remove the comment.

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## [0.1.5](https://github.com/acquire-project/acquire-driver-zarr/compare/v0.1.3...v0.1.4) - 2023-11-20
```
andy-sweet, line 8: I assume you're planning to tag the commit associated with merging this PR with v0.1.5? If so, looks good.

andy-sweet (approved):
Thanks!

PR #173, opened 2023-11-27
Use nlohman::json to validate (i.e., try-parse) external metadata.

aliddell:


andy-sweet (approved):
Looks good to me. Does this need a changelog entry?

src/zarr.cpp
```diff
@@ -18,14 +18,8 @@ validate_json(const char* str, size_t nbytes)
...
-    // Don't do full json validation here, but make sure it at least
-    // begins and ends with '{' and '}'
-    EXPECT(nbytes >= 3,
-           "nbytes (%d) is too small. Expected a null-terminated json string.",
-           (int)nbytes);
-    EXPECT(str[nbytes - 1] == '\0', "String must be null-terminated");
-    EXPECT(str[0] == '{', "json string must start with \'{\'");
-    EXPECT(str[nbytes - 2] == '}', "json string must end with \'}\'");
+    std::string metadata{ str, nbytes - 1 }; // don't include null terminator
+    json::parse(metadata);                   // throws if invalid
```
nclack, line 22: If we can use the iterator-based version of parse, it avoids a copy.

```suggestion
    json::parse(str,str+nbytes);                   // throws if invalid
```

nclack: I'm also for `allow_comments=true` here. I'd like the api to be permissive where possible so it's more usable by callers. The implementation should be strict about what it writes though. So if we're able to parse with comments and write clean (commentless) json, that seems best.

nclack (commented):
Small change suggested. Not critical. I'm not sure if I got the `end` part right.

src/zarr.cpp
aliddell: I'm fine to allow comments, even though it's highly irregular. I'm actually surprised the library supports JSON comments when JSON itself doesn't support comments. I don't think the user will be typically pasting in a JSON string with comments, but will rather be constructing them from a Python dict, so it won't make a difference one way or the other.

nclack (approved):

PR #188, opened 2023-12-08
Startup thread pool on start

aliddell:
- Updates acquire-core-libs, acquire-video-runtime, and acquire-driver common submodules to current main versions. (Closes #179. Closes #184).
- Starts the thread pool on `Zarr::start()`. Destroys the thread pool on `Zarr::stop()`.


nclack (approved):
👍

PR #197, opened 2023-12-15
Reset state on stop

aliddell:
- Fixes a couple of bugs in `Zarr::get()`
- Sets state in Zarr object from C API calls.
- Stores shard dims in units of chunks in Zarr V3 (previously px)
- Resets internal state on `Zarr::stop()`.

### DIFF
--- a/src/zarr.cpp
+++ b/src/zarr.cpp
@@ -127,12 +127,11 @@ DeviceState
 zarr_start(Storage* self_) noexcept
 {
     DeviceState state{ DeviceState_AwaitingConfiguration };
-
     try {
         CHECK(self_);
         auto* self = (zarr::Zarr*)self_;
         self->start();
-        state = DeviceState_Running;
+        state = self->state;
     } catch (const std::exception& exc) {
         LOGE("Exception: %s\n", exc.what());
     } catch (...) {
@@ -145,20 +144,18 @@ zarr_start(Storage* self_) noexcept
 DeviceState
 zarr_append(Storage* self_, const VideoFrame* frames, size_t* nbytes) noexcept
 {
-    DeviceState state;
+    DeviceState state{ DeviceState_AwaitingConfiguration };
     try {
         CHECK(self_);
         auto* self = (zarr::Zarr*)self_;
         *nbytes = self->append(frames, *nbytes);
-        state = DeviceState_Running;
+        state = self->state;
     } catch (const std::exception& exc) {
         *nbytes = 0;
         LOGE("Exception: %s\n", exc.what());
-        state = DeviceState_AwaitingConfiguration;
     } catch (...) {
         *nbytes = 0;
         LOGE("Exception: (unknown)");
-        state = DeviceState_AwaitingConfiguration;
     }
 
     return state;
@@ -172,8 +169,8 @@ zarr_stop(Storage* self_) noexcept
     try {
         CHECK(self_);
         auto* self = (zarr::Zarr*)self_;
-        CHECK(self->stop());
-        state = DeviceState_Armed;
+        CHECK(self->stop()); // state is set to DeviceState_Armed here
+        state = self->state;
     } catch (const std::exception& exc) {
         LOGE("Exception: %s\n", exc.what());
     } catch (...) {
@@ -320,6 +317,8 @@ average_two_frames(VideoFrame* dst, const VideoFrame* src)
 void
 zarr::Zarr::set(const StorageProperties* props)
 {
+    EXPECT(state != DeviceState_Running,
+           "Cannot set properties while running.");
     CHECK(props);
 
     StoragePropertyMetadata meta{};
@@ -424,6 +423,7 @@ zarr::Zarr::start()
       std::thread::hardware_concurrency(),
       [this](const std::string& err) { this->set_error(err); });
 
+    state = DeviceState_Running;
     error_ = false;
 }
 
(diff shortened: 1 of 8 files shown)


andy-sweet (commented):
I took a first pass through this, but not enough to approve/reject yet.

src/zarr.v3.cpp
```diff
@@ -34,13 +34,15 @@ void
...
-    // for now
-    ImageDims shard_dims = {
-        .cols = props.width,
-        .rows = props.height,
-    };
-    shard_dims_.push_back(shard_dims);
+    // tile shape (see reserve_image_shape) so we need to store the raw (i.e.,
+    // per chunk) values.
+    shard_dims_chunks_.clear();
+    shard_dims_chunks_.push_back({ std::clamp(props.width,
```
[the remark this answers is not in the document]
aliddell: Yes, exactly. It's to set up the case when it's supported.

src/zarr.cpp
```diff
@@ -127,12 +127,11 @@ DeviceState
 zarr_start(Storage* self_) noexcept
 {
     DeviceState state{ DeviceState_AwaitingConfiguration };
-
     try {
         CHECK(self_);
         auto* self = (zarr::Zarr*)self_;
         self->start();
-        state = DeviceState_Running;
+        state = self->state;
```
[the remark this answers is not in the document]
aliddell: There's not a great reason. `DeviceState_AwaitingConfiguration` is what was previously returned on failure, and not setting the state on, e.g., start and stop was the bug. Nowhere does the Zarr object ever set its own state to `DeviceState_AwaitingConfiguration`, except in the constructor. The contract around state changes could be better documented and enforced. That's how this managed to slip under the radar for so long.

OTOH we only check that the state is `DeviceState_Running` on `storage_start`, and it could be either `DeviceState_Armed` or `DeviceState_AwaitingConfiguration` on `storage_stop`, which means it should be fine to return current device state.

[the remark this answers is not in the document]
aliddell: Wait, there's a simpler reason. The `self_` pointer could be null. My point about the contract being sporadically enforced still stands though.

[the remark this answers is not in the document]
andy-sweet: My understanding of the intended state changes is not great, so this explanation makes sense in the context of this PR, so this is resolved (I.e. we should try to improve and document it in future work).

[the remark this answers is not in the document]
aliddell: 100%

src/zarr.cpp
```diff
@@ -587,6 +599,24 @@ zarr::Zarr::set_chunking(const ChunkingProps& props, const ChunkingMeta& meta)
...
+void
+zarr::Zarr::reset_state_()
+{
+    // reset state variables
+    writers_.clear();
+
+    // should be empty, but just in case
+    for (auto& [_, frame] : scaled_frames_) {
+        if (frame.has_value() && frame.value()) {
+            free(frame.value());
```
andy-sweet, line 611: The comment above provides some context, but this makes me ask the question who owns these video frames. I guess it's acquire's source? But `free` is called elsewhere in this implementation, so this isn't new, and thus is not a blocker for this PR.

src/zarr.cpp
```diff
@@ -587,6 +599,24 @@ zarr::Zarr::set_chunking(const ChunkingProps& props, const ChunkingMeta& meta)
     CHECK(planes_per_chunk_ > 0);
 }
 
+void
+zarr::Zarr::reset_state_()
```
andy-sweet, line 603: Nit: this is private, but `reset_state_` is not a great name given we have `Storage::state`, which this does not change.

Unfortunately, I don't have great ideas for different names. Maybe just `reset_` might be OK?

src/zarr.cpp
```diff
@@ -587,6 +599,24 @@ zarr::Zarr::set_chunking(const ChunkingProps& props, const ChunkingMeta& meta)
...
+    // should be empty, but just in case
+    for (auto& [_, frame] : scaled_frames_) {
+        if (frame.has_value() && frame.value()) {
+            free(frame.value());
+        }
+    }
+    scaled_frames_.clear();
+
+    error_ = false;
+    error_msg_.clear();
```
andy-sweet, line 617: `set_error` (just below) acquires a lock on `mutex_`, but this does not. Is that intended? There are other places where `error_` is mutated and the lock is not held, so this inconsistency may not be new. But it would be good to not add to it.

andy-sweet (changes requested):
My only real ask here is to change the name of `reset_state_` because it's unrelated to `Storage::state`. I have some other more open ended questions, but hopefully this looks close.

src/zarr.cpp
aliddell: The Zarr object owns them. They're created in `scale_image()` (on line 260 to be exact. Even the full-res images are copies of the frames from the source.

src/zarr.cpp
aliddell: Yeah, by this point the thread pool is shut down, destroyed even. Nobody else should be able to mutate it.

src/zarr.cpp
aliddell: Yeah, I didn't have any great ideas either. There's the "state" of it as a stateful object and there's the "device state" of it as a Storage device. Thinking of changing the comment in line 76 from `Setup` to `Setup / teardown`, so `teardown_` is an option. Again, not a great one.

src/zarr.cpp
andy-sweet: Oh thanks for clarifying. Resolved.

src/zarr.cpp
andy-sweet: Another alternative is to remove this method and inline the code into `stop` since that's the only place it's called. Then we don't have to think of any names :p

andy-sweet (approved):
I'm good with this.

src/zarr.cpp
aliddell: Fine by me.

PR #233, opened 2024-02-15
Support configurable append dimension

aliddell:
- Updates `acquire-common` submodule to latest.
- Adds support for nearly arbitrary dimension configurations.

The following need to be finished before this is merged:
- [ ] Update changelog to correct 0.10.0 release date.


andy-sweet (commented):
I need a lot more time with this before I'd be comfortable approving/rejecting, but thought I'd drop some minor comments I have so far.

tests/repeat-start.cpp
```diff
@@ -16,6 +16,22 @@
 namespace fs = std::filesystem;
 using json = nlohmann::json;
 
+namespace {
+void
+init_array(struct StorageDimension** data, size_t size)
+{
+    if (!*data) {
```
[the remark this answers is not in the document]
andy-sweet: As discussed live, at a minimum I'd want to have good documentation about the expected behavior of `storage_properties_dimensions_s.init`. Would also be interested to see the Rust implementation for the Python API.

README.md
```diff
@@ -52,21 +52,125 @@ os.environ["ZARR_V3_SHARDING"] = "1"
...
+    // the size of a chunk along this dimension
+    uint32_t chunk_size_px;
+
+    // the number of chunks in a shard along this dimension
+    uint32_t shard_size_chunks;
+};
+```
+
+Each of your output dimensions should have a corresponding `StorageDimension` struct.
+The order of these dimensions matters: the first dimension in the array will be the fastest-varying dimension as you
```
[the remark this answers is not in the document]
aliddell: Yeah. In light of yesterday's conversation this might be a good opportunity to reverse it.

src/zarr.cpp
```diff
@@ -77,6 +80,136 @@ validate_props(const StorageProperties* props)
     }
 }
 
+void
+validate_dimension(const zarr::Dimension& dim, bool is_append)
+{
+    if (!is_append) {
+        EXPECT(dim.array_size_px > 0, "Dimension array size must be positive.");
+        EXPECT(dim.chunk_size_px > 0 && dim.chunk_size_px <= dim.array_size_px,
```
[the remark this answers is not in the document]
aliddell: This is fixed.

tests/repeat-start.cpp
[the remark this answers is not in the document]
aliddell: See https://github.com/acquire-project/acquire-common/pull/38

src/writers/writer.cpp
```diff
@@ -8,255 +8,451 @@
...
-                      EXPECT(fs::create_directories(path),
-                             "Failed to create directory: %s",
-                             path.c_str());
+                  if (fs::exists(dirname)) {
+                      EXPECT(fs::is_directory(dirname),
+                             "'%s' exists but is not a directory.",
+                             dirname.c_str());
+                  } else if (success) {
+                      EXPECT(fs::create_directories(dirname),
+                             "Not creating directory '%s': another job failed.",
```
andy-sweet, line 240: How did you infer that another job failed in this case?

src/writers/writer.cpp
```diff
@@ -8,255 +8,451 @@
...
+                           dirname.string().c_str());
+                  err = buf;
+                  success = false;
+              }
+
+              latch.count_down();
+              return success;
+          });
+
+        dir_paths.push(dirname);
```
andy-sweet, line 266: The `front, pop, push` pattern is a bit strange because you're accessing a `queue` like a `vector`. Is there another component that needs this input need to be a `queue`?

src/writers/writer.cpp
```diff
@@ -8,255 +8,451 @@
...
+    }
+
+    latch.wait();
+
+    return success;
+}
+
+bool
+zarr::FileCreator::make_files_(std::queue<fs::path>& file_paths,
```
andy-sweet, line 275: Unlike `make_dirs_` this mutates `file_paths` by consuming the queue (i.e. we only `pop` here with no corresponding `push`). I think it's worth mentioning that distinction in the header docs for these methods.

andy-sweet (commented):
Took a high level look at `make_dirs_` and `make_files_` and don't see any big issues. I'm slightly confused about why the input to `make_dirs_` is a queue (and not a `vector`), but that's probably explained by its usage in another part of the code.

src/writers/writer.cpp
```diff
@@ -8,255 +8,451 @@
 
 namespace zarr = acquire::sink::zarr;
 
-/// DirectoryCreator
+namespace {
+/// Returns the index of the chunk in the lattice of chunks for the given frame
+/// and dimension.
+size_t
+chunk_lattice_index(size_t frame_id,
```
andy-sweet, line 15: Is `frame_id` still a good name here? Isn't this more like an index into the array in the given dimension (i.e. `dims[dimension_idx]`).

aliddell: It's using the frame id to compute and return that dimension.

src/writers/writer.cpp
aliddell: That's actually incorrect. Will fix.

src/writers/writer.cpp
aliddell: I'm not really doing random access. In order to build up the file paths, I'm doing what's essentially an `itertools.product()` over a number of arguments that I can't determine in advance. A queue is the best tool for this IMO.

src/writers/writer.cpp
aliddell: Can do

nclack (approved):

PR #240, opened 2024-04-18
Migrate filesystem writers to `Sink` concept

aliddell:
### Changed

- Calling `acquire_get_configuration()` with a Zarr storage device now returns a URI of the storage device, with file://
  scheme indicator and absolute path, assuming localhost.


src/zarr.cpp
```diff
@@ -362,9 +373,17 @@ zarr::Zarr::get(StorageProperties* props) const
     storage_properties_destroy(props);
 
     const std::string dataset_root = dataset_root_.string();
-    const char* filename =
-      dataset_root.empty() ? nullptr : dataset_root.c_str();
-    const size_t bytes_of_filename = filename ? dataset_root.size() + 1 : 0;
+    char* uri = nullptr;
```
jeskesen, line 376: Is there a reason for this to be a char* (instead of a std::string)?

jeskesen (approved):
Looks good.  Seems to be a pretty straightforward rename & reorganization.

src/zarr.cpp
aliddell: `storage_properties_init` wants a `char*` and I committed to the pointer too early. This is fixed.

Issue #237, opened 2024-04-15
Migrate filesystem writers to `Sink` concept


aliddell:
Depends on https://github.com/acquire-project/acquire-common/issues/42

aliddell:
Closed with #240

PR #244, opened 2024-04-25
Implement S3 sinks

aliddell:
Depends on #237.


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "acquire-common"]
 	path = acquire-common
 	url = ../acquire-common.git
+[submodule "aws-sdk-cpp"]
+	path = aws-sdk-cpp
+	url = git@github.com:aws/aws-sdk-cpp.git
```
jeskesen, line 6: Should this be a specific revision?

aliddell: I suppose we could put it there. It's already a specific rev that's committed to this repo, but it would be clearer to developers what tag that points to. I went down all kinds of routes, including removing the submodule, to get a good build/install solution, so if you had that in there and I took it out, I apologize.

PR #249, opened 2024-06-10
Enforce CXX standard in CMake

shlomnissan:
**Summary**

This PR enables `CMAKE_CXX_STANDARD_REQUIRED` to enforce C++20. CMake will raise
an error if a compatible compiler is not found.

(see inline comments for more.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,8 @@ include(cmake/simd.cmake)
 
 set(CMAKE_C_STANDARD 11)
 set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
 
 add_subdirectory(src)
 add_subdirectory(tests)
--- a/src/writers/blosc.compressor.hh
+++ b/src/writers/blosc.compressor.hh
@@ -1,10 +1,6 @@
 #ifndef H_ACQUIRE_ZARR_BLOSC_COMPRESSOR_V0
 #define H_ACQUIRE_ZARR_BLOSC_COMPRESSOR_V0
 
-#ifndef __cplusplus
-#error "This header requires C++20"
-#endif
-
 #include "blosc.h"
 #include "json.hpp"
 
--- a/src/writers/writer.hh
+++ b/src/writers/writer.hh
@@ -1,10 +1,6 @@
 #ifndef H_ACQUIRE_ZARR_WRITER_V0
 #define H_ACQUIRE_ZARR_WRITER_V0
 
-#ifndef __cplusplus
-#error "This header requires C++20"
-#endif
-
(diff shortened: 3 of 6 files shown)


CMakeLists.txt
```diff
@@ -10,8 +10,9 @@ include(cmake/install-prefix.cmake)
 include(cmake/wsl.cmake)
 include(cmake/simd.cmake)
 
-set(CMAKE_C_STANDARD 11)
```
shlomnissan, line 13: If this is a requirement for an external library, it should be defined there (another example of why a package manager would be very beneficial).

CMakeLists.txt
```diff
@@ -10,8 +10,9 @@ include(cmake/install-prefix.cmake)
...
 set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
shlomnissan, line 14: This flag is set to OFF by default, which means that if the compiler doesn't support the standard, it will fall back to the latest standard supported by the compiler. Turning this flag ON prevents the fallback behavior.

CMakeLists.txt
```diff
@@ -10,8 +10,9 @@ include(cmake/install-prefix.cmake)
...
+set(CMAKE_CXX_EXTENSIONS OFF)
```
shlomnissan, line 15: This is a minor change. Modifying the flag is used to set the language standard (e.g. -std=c++11 rather than -std=gnu++11) to increase compatibility across compilers.

shlomnissan, line 15: This is a minor change. Modifying the flag sets the language standard (e.g. -std=c++11 rather than -std=gnu++11) to increase compatibility across compilers.

CMakeLists.txt
```diff
@@ -12,6 +12,8 @@ include(cmake/simd.cmake)
 
 set(CMAKE_C_STANDARD 11)
```
shlomnissan, line 13: This feels out of place. If this is a requirement for another library, shouldn't it be defined there?

aliddell: What makes it feel out of place?

aliddell (approved):
I like it.

CMakeLists.txt
shlomnissan: It seems that the codebase is using C++20. If we have dependencies that use C11, we should probably have target-specific specifications instead of global ones — either by setting the requirements in the dependency build configuration, or by explicitly setting the specification for those targets:

```cpp
add_library(MyLibrary SHARED my_c_library.c)
set_target_properties(MyLibrary PROPERTIES
    C_STANDARD 11
    C_STANDARD_REQUIRED ON
)
```

This ensures consistency and prevents compatibility issues. It also improves clarity by documenting where specific standards are used, making it easier to understand the build configuration.

shlomnissan:
Feel free to merge this PR. I don't have write access.

PR #254, opened 2024-06-16
Add vcpkg and remove hardcoded dependencies

shlomnissan:
**Summary**

This PR adds vcpkg configurations for dependency management and removes
hardcoded dependencies from the codebase. It also adds vcpkg to our GitHub
Actions workflow and uses it to automatically install our dependencies.

**Notes**
- By integrating a package manager, we significantly simplify dependency management, improve our productivity, and promote consistency across environments.
- In this configuration, there's no need to add vcpkg as a submodule to the project. Instead, individual users should install vcpkg in their home directory and set up the necessary environment variables (following this [guide](https://learn.microsoft.com/en-us/vcpkg/get_started/get-started?pivots=shell-bash).)
- When we're ready to add [minio](https://vcpkg.io/en/package/minio-cpp), all we need to do is update `vcpkg.json` in the root folder.
- I will update the README file with installation instructions and remove the hardcoded `json.hpp` library from the tests folder in separate PRs.

**Test Plan**
- All existing tests have passed.


.github/workflows/test_pr.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - main
-  pull_request_target:
+  pull_request:
```
shlomnissan, line 7: It's strange that `pull_request_target` is still running. I assume this is expected? Do we typically revert this line before squashing the PR?

.github/workflows/test_pr.yml
```diff
@@ -20,6 +20,13 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ "windows-latest", "ubuntu-latest", "macos-latest" ]
+        include:
+          - platform: "windows-latest"
+            vcpkg_triplet: "x64-windows-static"
```
shlomnissan, line 25: I had to manually select compiler triplets because of the way blosc is configured on vcpkg. The type of target library (static/shared) is encoded in the name of the library (i.e., blosc_static). Ubuntu and MacOS, by default, will compile blosc as a static library and Windows as a shared library. By selecting the triplet for Windows manually (`x64-windows-static`), we're ensuring that the output is consistent.

I may submit a patch to vcpkg. This is annoying.

CMakeLists.txt
```diff
@@ -3,6 +3,9 @@ project(acquire-driver-zarr)
 cmake_policy(SET CMP0079 NEW) # allows use with targets in other directories
 enable_testing()
 
+find_package(nlohmann_json CONFIG REQUIRED)
+find_package(blosc CONFIG REQUIRED)
```
shlomnissan, line 7: The `CONFIG` keyword is used to specify that the package should be found using a configuration file. CONFIG mode aligns with modern CMake where libraries and packages are expected to provide their own configuration files.

.github/workflows/test_pr.yml
aliddell: Yeah, `pull_request_target` will still run whatever workflow file was committed to `main`.

.github/workflows/test_pr.yml
aliddell: That's interesting, do you need triplets for all three platforms?

.github/workflows/test_pr.yml
```diff
@@ -43,8 +50,19 @@ jobs:
...
         shell: bash
 
+      - name: Install vcpkg
+        run: |
+          git clone https://github.com/microsoft/vcpkg.git
+          cd vcpkg && ./bootstrap-vcpkg.sh
+          echo "VCPKG_ROOT=${{github.workspace}}/vcpkg" >> $GITHUB_ENV
+          echo "${{github.workspace}}/vcpkg" >> $GITHUB_PATH
+          ./vcpkg integrate install
+        shell: bash
```
aliddell, line 60: Will we recommend developers install vcpkg on their own then?

CMakeLists.txt
aliddell, line 7: We like to keep the base CMakeLists.txt pretty clean. As the list of dependencies grows, this could become pretty cluttered. Would you mind putting this into the `cmake` folder, in `dependencies.cmake` or something like that?

aliddell (changes requested):
I'd ask for two things:

1. Put `find_package` statements in a separate file to `include` in the base.
2. Add a note about installing vcpkg in the readme somewhere.

.github/workflows/test_pr.yml
shlomnissan: Yes, this is the recommended approach - I plan to add installation instructions to the README file. The advantage of using a preset instead of hard-coding the toolchain is that it allows developers to install dependencies in their preferred way.

CMakeLists.txt
shlomnissan: - For small to medium projects, I typically choose to place `find_package` commands directly in the main CMakeLists.txt file. This makes it easier to understand the dependencies at a glance, slightly reducing the overhead of understanding the project structure.
- Creating separate CMake modules can help keep the main CMakeLists.txt file tidy. In the past, I created them mainly for reusability. However, I don't believe this is a particularly good example of a reusable component.
- Ideally, we keep our dependency list small.

I don't think it's a big deal either way, so if you feel strongly about this, I'm happy to move it to a separate module.

shlomnissan:
> 1. Put find_package statements in a separate file to include in the base.

I've responded to the comment. Let me know your thoughts.

> 2. Add a note about installing vcpkg in the readme somewhere.

Yes, this was included in my PR summary notes. I initially thought of adding it to a subsequent PR, but it makes sense to update the README file here. I will update it later today.

shlomnissan:
@aliddell

I added vpkg installation instructions to the README file.

aliddell (approved):

aliddell:
Don't forget to change back to `pull_request_target`!

PR #262, opened 2024-07-05
Misc fixes

aliddell:
- Use the x64-osx triplet for build and release jobs on Mac (see error [here](https://github.com/acquire-project/acquire-driver-zarr/actions/runs/9765062597/job/26958778874))
- Link nlohmann_json::nlohmann_json in examples
- Fix array size bug in downsampling

### DIFF
--- a/examples/no-striping.cpp
+++ b/examples/no-striping.cpp
@@ -12,7 +12,7 @@
 #include <fstream>
 #include <stdexcept>
 
-#include "tests/json.hpp"
+#include <nlohmann/json.hpp>
 
 namespace fs = std::filesystem;
 using json = nlohmann::json;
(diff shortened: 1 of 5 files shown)


examples/no-striping.cpp
```diff
@@ -12,7 +12,7 @@
 #include <fstream>
 #include <stdexcept>
 
-#include "tests/json.hpp"
+#include <nlohmann/json.hpp>
```
shlomnissan, line 15: I think we can delete `tests/json.hpp` at this point, right?

shlomnissan (approved):
LGTM :+1:

examples/no-striping.cpp
aliddell: I think you did this already in an earlier PR.

PR #264, opened 2024-07-05
238, part 3: Separate common classes and utilities. Fix includes. Move tests.

aliddell:
- Moves classes and functions from common.hh/common.cpp to separate files in src/common
- Updates header includes accordingly
- Move shard index tests to common/utilities.cpp
- Adds a thread pool unit test.

Depends on #262.


src/common/thread.pool.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+namespace zarr = acquire::sink::zarr;
+namespace common = zarr::common;
+
+common::ThreadPool::ThreadPool(size_t n_threads,
+                               std::function<void(const std::string&)> err)
+  : error_handler_{ err }
+  , is_accepting_jobs_{ true }
+{
+    n_threads = std::clamp(
```
shlomnissan, line 12: 1. Explicit `static_cast` is preferred in modern C++ because it clearly indicates the type of conversion.
2. Use `size_t` consistently instead of mixing `unsigned`.
```cpp
n_threads = std::clamp(n_threads, static_cast<std::size_t>(1),
            std::max(std::thread::hardware_concurrency(), static_cast<std::size_t>(1)));
```

src/common/thread.pool.cpp
```diff
@@ -0,0 +1,155 @@
...
+        std::scoped_lock lock(jobs_mutex_);
+        while (!jobs_.empty()) {
+            jobs_.pop();
+        }
+    }
+
+    await_stop();
+}
+void
+common::ThreadPool::push_to_job_queue(JobT&& job)
```
shlomnissan, line 33: I think that `lock.unlock()` is called unnecessarily before notifying `cv_`, which can lead to a race condition if another thread is waiting on the condition variable.

```cpp
{
    std::unique_lock lock(jobs_mutex_);
    CHECK(is_accepting_jobs_);
    
    jobs_.push(std::move(job));
}

cv_.notify_one();
```

src/writers/writer.hh
```diff
@@ -4,7 +4,8 @@
 #include "platform.h"
 #include "device/props/components.h"
 
-#include "../common.hh"
+#include "common/dimension.hh"
```
shlomnissan, line 7: :+1: getting rid of relative paths!

src/common/utilities.cpp
```diff
@@ -0,0 +1,495 @@
...
+    }
+
+    return n_bytes;
+}
+
+const char*
+common::sample_type_to_dtype(SampleType t)
+
+{
+    static const char* table[] = { "u1", "u2", "i1", "i2",
```
shlomnissan, line 173: This type of logic feels very fragile to me. I think it would be better to have a complete switch statement around the values of `SampleType` and return their `dtype` counterparts.

src/common/utilities.cpp
```diff
@@ -0,0 +1,495 @@
...
+                                   "f4", "u2", "u2", "u2" };
+    if (t < countof(table)) {
+        return table[t];
+    } else {
+        throw std::runtime_error("Invalid sample type.");
+    }
+}
+
+const char*
+common::sample_type_to_string(SampleType t) noexcept
```
shlomnissan, line 183: Why does this function return a string if there’s an error (coupled with `noexcept`) and the previous one throws an exception?

src/common/utilities.cpp
```diff
@@ -0,0 +1,495 @@
...
+size_t
+common::align_up(size_t n, size_t align)
+{
+    EXPECT(align > 0, "Alignment must be greater than zero.");
+    return align * ((n + align - 1) / align);
+}
+
+std::vector<std::string>
+common::split_uri(const std::string& uri)
+{
```
shlomnissan, line 203: Do we want to handle empty URIs?

src/common/utilities.cpp
```diff
@@ -0,0 +1,495 @@
...
+    const char delim = '/';
+
+    size_t begin = 0;
+    auto end = uri.find_first_of(delim);
+
+    std::vector<std::string> out;
+    while (end <= std::string::npos) {
```
shlomnissan, line 210: The condition `while (end <= std::string::npos)` is a bit unconventional. It would be clearer to use `while (end != std::string::npos)` instead.

src/common/utilities.cpp
```diff
@@ -0,0 +1,495 @@
...
+        if (end > begin) {
+            out.emplace_back(uri.substr(begin, end - begin));
+        }
+
+        if (end == std::string::npos) {
```
shlomnissan, line 215: Is this check necessary? If `end == std::string::pos` the loop will be terminated anyway, right?

shlomnissan (approved):
Most of my comments are nits, so I'm going to approve the PR. I would like to see some of these changes implemented before you merge it, but nothing is blocking on my end.

src/common/dimension.hh
```diff
@@ -0,0 +1,26 @@
...
+
+#include "device/props/storage.h"
+
+#include <string>
+
+namespace acquire::sink::zarr {
+struct Dimension
+{
+  public:
+    explicit Dimension(const std::string& name,
```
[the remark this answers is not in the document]
aliddell: > The caller can initialize the object this way:
>
> ```
> const auto d = Dimension {{
>   .name = "name",
>   .kind = ...,
>   .array_size_pix = ...
>   ...
> }};
> ```

Not exactly. `StorageDimension::name` is a `struct String` which has the following interface:

```c
struct String
{
    char* str;
    size_t nbytes;
    uint8_t is_ref;
};
```

So there's a little more manual intervention than your example.

Anyway, there are two constructors because there are two modes of initialization. The first is from a `struct StorageDimension` in `Zarr::set_dimensions_()`, and the second is in `downsample(...)` in `writer.cpp` (line 132), taking a `Dimension` and downsampling it.

src/common/dimension.cpp
```diff
@@ -0,0 +1,32 @@
+#include "dimension.hh"
+
+#include <cmath>
```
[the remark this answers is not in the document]
aliddell: We don't. I think it's an artifact of CLion's Refactor tool. I can also remove the implementation file.

src/common/dimension.cpp
```diff
@@ -0,0 +1,32 @@
...
+                           uint32_t chunk_size_px,
+                           uint32_t shard_size_chunks)
+  : name{ name }
+  , kind{ kind }
+  , array_size_px{ array_size_px }
+  , chunk_size_px{ chunk_size_px }
+  , shard_size_chunks{ shard_size_chunks }
+{
+    EXPECT(kind < DimensionTypeCount, "Invalid dimension type.");
+    EXPECT(!name.empty(), "Dimension name cannot be empty.");
```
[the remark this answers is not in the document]
aliddell: Trimming is a good idea. Added.

src/common/thread.pool.hh
```diff
@@ -0,0 +1,50 @@
...
+    std::function<void(const std::string&)> error_handler_;
+
+    std::vector<std::thread> threads_;
+    mutable std::mutex jobs_mutex_;
+    std::condition_variable cv_;
+    std::queue<JobT> jobs_;
+
+    std::atomic<bool> is_accepting_jobs_;
+
+    std::optional<ThreadPool::JobT> pop_from_job_queue_() noexcept;
```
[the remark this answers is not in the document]
aliddell: We tend to use `[[nodiscard]]` for error-indicating return values (`bool write()`, for example).

src/common/utilities.cpp
aliddell: The intended use is different. `sample_type_to_dtype` is intended to write to Zarr metadata, so if the pixel type is unsupported, the program had better not silently write anything like "unrecognized pixel type". On the other hand, `sample_type_to_string` is simply used for logging, so the correct thing to do when pass an unrecognized `SampleType` is log that.

src/common/thread.pool.cpp
aliddell: What we have happening here is `std::thread::hardware_concurrency()` returning `unsigned int` (usually uint32), which isn't a `size_t` in general. The only reason we're using `size_t` here at all is because that's what the constructor asks for. It would actually be better to modify it constructor to take `unsigned int`, so we don't have to do so much casting.

src/common/utilities.cpp
aliddell: Do you mean do we want to accept them and return an empty vector if so? Why not?

src/common/utilities.cpp
aliddell: No, not sure what that's doing there.

src/common/thread.pool.cpp
aliddell: You may be right, but it's always been tough for me to reason about these things. My understanding of what's going on at this moment is that all threads are either going to be executing a job or waiting on the condition variable. Those threads that are executing will then cycle back to the beginning of the loop and try to acquire a lock on the mutex, which they immediately relinquish when they start waiting on the condition variable. The potential race condition I see here is that

1. the thread pool unlocks the lock
2. a worker thread acquires the lock
3. the thread pool calls `cv_.notify_one()`
4. the worker starts waiting on the condition variable after (3).

This should be fine, right? Some other thread can pick up the notify. This is of course assuming there's more than one worker thread.

src/common/thread.pool.hh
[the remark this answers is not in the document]
shlomnissan: That makes sense. Using `[[nodiscard]]` for error handling and resource management is the canonical use case. I personally add it to most functions that return a value unless it's okay for the client to ignore it. A good example of why this broader approach is valuable is `std::vector::empty()`. If you're new to C++, you might mistakenly think it clears the vector. Since the standard library marked it as `[[nodiscard]]`, the intent is much clearer, and it can be enforced by static analyzers.

src/common/thread.pool.cpp
shlomnissan: You can use unsigned literals (i.e., `1u`) instead of creating a trivial variable.

src/common/thread.pool.cpp
shlomnissan: It sounds about right, but I think it’s based on assumptions that may not always hold true. When you notify the condition variable after releasing the lock, you risk a situation where:

1. The thread pool unlocks the mutex.
2. A worker thread acquires the lock, but it doesn't see the new job yet because it hasn’t started waiting on the condition variable.
3. The thread pool calls `cv_.notify_one()`.
4. The worker thread then starts waiting on the condition variable after the notify, missing the notification.

I think this can lead to inefficiencies where the notified thread doesn't see the job that was just pushed, causing it to go back to waiting unnecessarily, while another worker thread might pick up the job eventually. It's generally best practice to hold the lock while notifying to ensure that the notified thread immediately sees the state change.

src/common/utilities.cpp
shlomnissan: Sorry, my comment wasn't very clear. I meant do we want to add error handling for empty URIs or are we considering empty strings as valid input to this function.

aliddell: I'm leaving error handling to the caller here.

PR #268, opened 2024-07-08
Update reported datatypes to conform with specs

aliddell:
- [Zarr V2](https://zarr-specs.readthedocs.io/en/latest/v2/v2.0.html#data-type-encoding)
- [Zarr V3](https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#core-data-types)

zarr-python had been reading our datatypes without complaining. Testing by @chris-delg has revealed that TensorStore will not.

## Changes

- Link against nlohman::json in examples builds (orthogonal to the issue, but a light lift).
- Remove `sample_type_to_dtype()` from common.{h,cpp}
- Implement version-specific `sample_type_to_dtype()`
- Update tests to check data type is correct.

### DIFF
--- a/src/zarr.v2.cpp
+++ b/src/zarr.v2.cpp
@@ -4,6 +4,8 @@
 
 #include "nlohmann/json.hpp"
 
+#include <bit>
+
 namespace zarr = acquire::sink::zarr;
 
 namespace {
@@ -21,6 +23,21 @@ compressed_zarr_v2_init()
         LOGE("Exception: (unknown)");
     }
     return nullptr;
+}
+
+std::string
+sample_type_to_dtype(SampleType t)
+
+{
+    const std::string dtype_prefix =
+      std::endian::native == std::endian::big ? ">" : "<";
+
+    std::string table[] = { "u1", "u2", "i1", "i2", "f4", "u2", "u2", "u2" };
+    if (t < countof(table)) {
+        return dtype_prefix + table[t];
+    } else {
+        throw std::runtime_error("Invalid sample type.");
+    }
 }
 } // end ::{anonymous} namespace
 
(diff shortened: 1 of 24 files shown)


src/zarr.v2.cpp
```diff
@@ -22,6 +24,21 @@ compressed_zarr_v2_init()
...
 }
+
+std::string
+sample_type_to_dtype(SampleType t)
+
+{
+    const std::string dtype_prefix =
+      std::endian::native == std::endian::big ? ">" : "<";
+
+    std::string table[] = { "u1", "u2", "i1", "i2", "f4", "u2", "u2", "u2" };
```
shlomnissan, line 35: It seems that the size of the table is based on the assumption that `SampleType` enum values directly correspond to indices in the table array. If that’s the case, this type of logic is very fragile. I would use a switch statement that covers all the cases of `SampleType`.

```cpp
std::string sample_type_to_dtype(SampleType t) {
    const auto prefix = std::endian::native == std::endian::big ? ">" : "<";

    switch (t) {
        case SampleType_u8:
            return prefix + "u1";
        case SampleType_u16:
            return prefix + "u2";
        case SampleType_i8:
            return prefix + "i1";
        case SampleType_i16:
            return prefix + "i2";
        case SampleType_f32:
            return prefix + "f4";
        case SampleType_u10:
        case SampleType_u12:
        case SampleType_u14:
            return prefix + "u2";
        default:
            throw std::runtime_error("Invalid SampleType: " + std::to_string(static_cast<int>(t)));
    }
}
```

1. The switch statements clearly express intent of handling different cases based on the enum value.
2. Switch statements are often checked at compile time to ensure all enum values are covered.
3. Adding new cases or modifying existing ones will be straightforward.
4. Switch typically compile to jump tables or direct comparisons which may result in better performance.

I think this applies to `std::string sample_type_to_dtype(SampleType t)` as well.

aliddell: I think this is fair.

shlomnissan (approved):
LGTM

PR #269, opened 2024-07-08
Support multiscale arrays in Zarr V3 with provisional metadata

aliddell:
This PR uses the same OME metadata as our Zarr V2 writer to express multiscale semantics. The following changes were made:

- A protected method, `Zarr::make_multiscale_metadata_()`, was implemented by moving the code from `ZarrV2::write_group_metadata_()` so that it could be made accessible to both `ZarrV2` and `ZarrV3` instances.
  - Call `make_multiscale_metadata_()` from both overrides of `write_group_metadata_()`.
- Update the metadata returned by `Zarr::get_meta()` to reflect that both types of Zarr store support multiscale.

Basic functionality is verified in the `write-zarr-v3-raw-multiscale.cpp` test file. The test scheme is the same as in `write-zarr-v2-raw-multiscale.cpp`, but the paths are changed to reflect the different layout.


src/zarr.cpp
```diff
@@ -709,6 +700,104 @@ zarr::Zarr::write_mutable_metadata_() const
...
+        multiscales[0]["datasets"] = {
+            {
+              { "path", "0" },
+              { "coordinateTransformations",
+                {
+                  {
+                    { "type", "scale" },
+                    { "scale", scales },
+                  },
+                } },
```
jeskesen, line 759: trivial, but this indenting is confusing

src/zarr.cpp
```diff
@@ -709,6 +700,104 @@ zarr::Zarr::write_mutable_metadata_() const
...
+
+            multiscales[0]["datasets"].push_back({
+              { "path", std::to_string(i) },
+              { "coordinateTransformations",
+                {
+                  {
+                    { "type", "scale" },
+                    { "scale", scales },
+                  },
+                } },
```
jeskesen, line 780: trivial, but this indenting is confusing

jeskesen (approved):
Looks good to me, other than a couple of nitpicky indentation things.

PR #270, opened 2024-07-10
Remove Automerge job

aliddell:
Since we don't have as many submodules as we used to, and we don't update it as frequently as we used to, handling Dependabot PRs manually is not as much of a chore as it used to be. Consequently, we don't need Automerge anymore, and since we don't need the secret, we can change our trigger back to `pull_request`.

- Remove automerge job
- Change test trigger from `pull_request_target` back to `pull_request`

### DIFF
--- a/.github/workflows/test_pr.yml
+++ b/.github/workflows/test_pr.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


shlomnissan (approved):
LGTM

PR #279, opened 2024-07-16
Add an append method that takes frame data, size, and shape

aliddell:
I updated `Zarr::append()` to call the new `Zarr::append_frame()`. Multiscale logic goes from `append()` into `append_frame()`, so I needed to retool `Zarr::write_multiscale_frames_()` to accept `uint8_t*` and `size_t` instead of `VideoFrame`.

I also updated `Writer::write()` to both take `uint8_t*` and `size_t` instead of `VideoFrame`, and also return a `size_t` of the number of bytes written in anticipation of establishing a contract around writing (separate PR). Because of this change, I had to update the unit tests around `{Writer,ZarrV2Writer,ZarrV3Writer}::write()`, and I took the opportunity to define constant values instead of hardcoding numbers. The tests are otherwise the same.

Also the unit tests are slightly reordered for aesthetic reasons.

### DIFF
--- a/src/writers/writer.cpp
+++ b/src/writers/writer.cpp
@@ -188,21 +188,19 @@ zarr::Writer::Writer(const WriterConfig& config,
     data_root_ = config_.data_root;
 }
 
-bool
-zarr::Writer::write(const VideoFrame* frame)
+size_t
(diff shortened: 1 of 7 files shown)


jeskesen (approved):
looks good to me.

PR #280, opened 2024-07-16
Mult-language API (C++ & Python)

jeskesen:
C++ & Python API.  

The C++ API has been written such that all internal data structures are compiled & statically linked into the dynamic library,  such that the artifacts are only a single header file + the dynamic library itself.  To do this, it has the external-facing class which does not reference any Acquire internals other than a forward-declared implementation class, which is not exposed and contains all the actual data structures. 

The python API simply inherits the C++ API and implements anything python-specific using pybind11.  Since pybind11 uses getters & setters to define properties of its classes, this has driven the C++ API to do the same. 

Note that these external APIs are not built by default and you have to "opt-in" using CMake variables.


api/cpp/include/acquire-zarr/acquire-zarr.hh
```diff
@@ -0,0 +1,242 @@
...
+    DTYPE_INT16,
+    DTYPE_FLOAT32,
+    DTYPE_UNIMPLEMENTED
+} ;
+
+class AcquireZarrWriter 
+{
+  public:
+    AcquireZarrWriter() { create_impl();}
+    ~AcquireZarrWriter() = default;
```
shlomnissan, line 31: @jeskesen I think the reason you can't use a unique pointer is that your destructor is defined here. This is equivalent to defining an inline destructor in the header file. The automatic deletion of the unique pointer happens in the destructor, but in this case, the compiler will raise an error because the size of the underlying type the pointer points to is unknown. If that's the underlying issue, the solution would be declaring the destructor in the header file and defining it in the implementation file.

```cpp
~AcquireZarrWriter(); // remove = default
```

And define it in the implementation file:
```cpp
AcquireZarrWriter::~AcquireZarrWriter() = default;
```
Try it out and let me know if it works. It’s been annoying me for some time now.

PR #286, opened 2024-08-08
Fix inaccurate reporting of append dimension size in metadata.

aliddell:
I noticed an issue where the shape of the array in the final dimension was always reported as the number of frames acquired, which, in the case of nontrivial internal dimensions, is just wrong. Opening up such a dataset in napari will confirm.


jeskesen (approved):
Approving because my comments are more a stylistic suggestion than anything, and shouldn't be blocking. 

I will also note that the fact that had to copy/paste the same suggestion suggests copy/pasted code (which IMHO should be a avoided).

jeskesen (approved):
Approving because my comments are more a stylistic suggestion than anything, and shouldn't be blocking. 

I will also note that the fact that had to copy/paste the same suggestion suggests copy/pasted code (which IMHO should be a avoided).

PR #287, opened 2024-08-15
Make the combination of get and set idempotent

aliddell:
- Use `std::string` when generating paths from `StorageProperties::uri`.
- Removes a redundant check of directory existence.
- Fixes a bug wherein web URIs are prepended with "file://"

### DIFF
--- a/src/zarr.cpp
+++ b/src/zarr.cpp
@@ -392,10 +392,13 @@ zarr::Zarr::get(StorageProperties* props) const
     storage_properties_destroy(props);
 
     std::string uri;
-    if (!dataset_root_.empty()) {
+    if (common::is_web_uri(dataset_root_)) {
+        uri = dataset_root_;
+    } else if (!dataset_root_.empty()) {
         fs::path dataset_root_abs = fs::absolute(dataset_root_);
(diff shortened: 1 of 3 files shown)


aliddell:
When you `get` the storage properties from the Zarr driver, if it's a I can't work out exactly why I was seeing this behavior when using `strcmp` but previously one would `get` properties and see a "file://" prefix. Taking that value and re`set`ting it would fail to cut out the "file://" prefix, using `strcmp` anyway. I can't work out exactly why `strcmp` was failing to do so, but that's what I noticed when testing. This fixes it and adds a test for it.`

src/zarr.cpp
```diff
@@ -68,19 +70,16 @@ validate_props(const StorageProperties* props)
         CHECK(tokens.size() > 2); // http://endpoint/bucket
     } else {
         const fs::path path = as_path(*props);
-        fs::path parent_path = path.parent_path().string();
+        LOG("path: '%s'", path.string().c_str());
```
aliddell, line 73: Don't merge while this line is still here.

nclack (approved):
lgtm. don't forget about removing that log line before merging.

PR #291, opened 2024-08-30
Define the Zarr streaming API

aliddell:
This defines the Zarr streaming API in zarr.h. It also moves driver tests to tests/driver and adds logger code and code for setting parameters on the Zarr stream.

The bulk of the changed files are just moved.


src/include/zarr.h
```diff
@@ -0,0 +1,550 @@
...
+    /***************************************************************************
+     * Functions for getting parameters on the Zarr stream.
+     *
+     * These functions return the value of the specified parameter.
+     * If the Zarr stream is NULL, the functions return NULL or 0.
+     **************************************************************************/
+
+    ZarrVersion ZarrStream_get_version(const ZarrStream* stream);
+
+    const char* ZarrStream_get_store_path(const ZarrStream* stream);
```
shlomnissan, line 486: Are these getters for the same underlying object, but instead of passing the settings object, you passed the stream? If that's the case, why not use one or the other? I'm still trying to understand the benefit of exposing every parameter as an accessor method, so I'm surprised to see another set of accessor functions.

I may need to learn more about this API, but currently it feels like a lot of unnecessary flexibility (and maybe a violation of the principle of least exposure). I might be misunderstanding the library's purpose, but I'm wondering: if I've already provided this information, why are accessor methods needed to retrieve it? Wouldn't returning an instance of the settings object be more straightforward?

shlomnissan (commented):
This is a partial review. It covers the main `CMakeLists.txt` file and the public interface `zarr.h`. I will continue reviewing the C++ code separately.

src/internal/logger.cpp
```diff
@@ -0,0 +1,81 @@
+#include "logger.hh"
```
shlomnissan, line 1: Don't we already have a logger in the `common` repo? Why do we need another one? If there are certain benefits to this logger, should we make it our primary logger?

src/internal/logger.cpp
```diff
@@ -0,0 +1,81 @@
...
+    return current_level;
+}
+
+std::string
+Logger::log(LogLevel level,
+            const char* file,
+            int line,
+            const char* func,
+            const char* format,
+            ...)
```
shlomnissan, line 29: C++ offers better alternatives for handling variable arguments, such as variadic templates and initializer lists, which are type-safe and more flexible. I think it should work with `__VA_ARGS__`.

src/internal/logger.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+LogLevel
+Logger::get_log_level()
+{
+    return current_level;
+}
+
+std::string
+Logger::log(LogLevel level,
```
shlomnissan, line 24: I don't think this code is thread-safe. If multiple threads log messages simultaneously, the output might become interleaved or corrupted.

src/internal/logger.cpp
```diff
@@ -0,0 +1,81 @@
...
+            const char* file,
```
shlomnissan, line 25: Can we use `std::string_view` instead of `const char*`?

src/internal/logger.cpp
```diff
@@ -0,0 +1,81 @@
...
+    // Get filename without path
+    std::filesystem::path filepath(file);
+    std::string filename = filepath.filename().string();
+
+    // Output timestamp, log level, filename
+    *stream << std::put_time(std::localtime(&time), "%Y-%m-%d %H:%M:%S") << '.'
+            << std::setfill('0') << std::setw(3) << ms.count() << " " << prefix
+            << filename << ":" << line << " " << func << ": ";
+
+    char buffer[1024];
```
shlomnissan, line 74: - If the formatted log message exceeds this size, wouldn't it cause a buffer overflow?
- Why not use `std::string`, which is the expected return type anyway?

src/internal/logger.hh
```diff
@@ -0,0 +1,38 @@
+#include "zarr.h"
```
shlomnissan, line 1: This design seems fragile. We're incorporating a C API wrapper into a logger class that should be generalized for accessing properties—properties that ought to be defined here initially. Are there any alternative approaches we could consider?

src/internal/logger.cpp
```diff
@@ -0,0 +1,81 @@
...
+            stream = &std::cerr;
+            break;
+    }
+
+    // Get current time
+    auto now = std::chrono::system_clock::now();
+    auto time = std::chrono::system_clock::to_time_t(now);
+    auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(
+                now.time_since_epoch()) %
+              1000;
```
shlomnissan, line 63: Nit: I think we can organize this code better. It would be an improvement to have a private member function that returns the current timestamp as a string, which would encapsulate this code, as well as `std::put_time`.

src/internal/logger.hh
```diff
@@ -0,0 +1,38 @@
...
+    Logger::log(LogLevel_Error, __FILE__, __LINE__, __func__, __VA_ARGS__)
+
+#define EXPECT(e, ...)                                                         \
+    do {                                                                       \
+        if (!(e)) {                                                            \
+            const std::string __err = LOG_ERROR(__VA_ARGS__);                  \
+            throw std::runtime_error(__err);                                   \
+        }                                                                      \
+    } while (0)
+#define CHECK(e) EXPECT(e, "Expression evaluated as false:\n\t%s", #e)
```
shlomnissan, line 38: I feel like all these macros can be replaced with static member functions, for example:
```cpp
template <typename... Args>
static void debug(const char* format, Args... args) {
   log(LogLevel_Debug, __FILE__, __LINE__, __func__, format, args...);
}
```
This approach improves type checking (which you get from variadic templates), and it avoids all the issues that come with macro expansions.

shlomnissan (commented):
This is another partial review that focuses on logging.

src/CMakeLists.txt
```diff
@@ -1,9 +1,54 @@
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+
+####### Acquire Zarr Streaming Library #######
+
+set(tgt acquire-zarr)
```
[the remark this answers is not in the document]
aliddell: At some point, the streaming code is going to go into its own repo, but this is probably a good opportunity to separate them.

src/include/zarr.h
```diff
@@ -0,0 +1,550 @@
...
+     **************************************************************************/
+    typedef enum
+    {
+        LogLevel_Debug,
+        LogLevel_Info,
+        LogLevel_Warning,
+        LogLevel_Error,
+        LogLevel_None,
+        LogLevelCount
+    } LogLevel;
```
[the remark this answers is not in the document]
aliddell: I have further down in the file a function to set the log level, so the user needs to know about them. I can move this and other enums that are also used internally to a separate header(s) if that's your objection.

src/include/zarr.h
```diff
@@ -0,0 +1,550 @@
...
+     * ZarrCompressionCodec_None value should only be used when not compressing.
+     * If the compressor is set to ZarrCompressor_None, the codec is ignored.
+     **************************************************************************/
+
+    typedef enum
+    {
+        ZarrCompressor_None = 0,
+        ZarrCompressor_Blosc1,
+        ZarrCompressor_Blosc2, /* Blosc2 is not yet supported */
+        ZarrCompressor_Zstd,   /* Zstd is not yet supported */
```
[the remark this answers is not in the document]
aliddell: I added them because I'd like to support them, but you're right, they don't belong here right now.

src/include/zarr.h
```diff
@@ -0,0 +1,550 @@
...
+
+    /** @brief Return a pointer to a Zarr stream settings struct. */
+    ZarrStreamSettings* ZarrStreamSettings_create();
+
+    /** @brief Destroy a Zarr stream settings struct. Consumes the pointer */
+    void ZarrStreamSettings_destroy(ZarrStreamSettings* settings);
+
+    /** @brief Make a copy of the Zarr stream settings struct. */
+    ZarrStreamSettings* ZarrStreamSettings_copy(
+      const ZarrStreamSettings* settings);
```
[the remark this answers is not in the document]
aliddell: I try to where they're obvious.

src/internal/stream.settings.cpp
```diff
@@ -0,0 +1,539 @@
...
+                                        size_t bytes_of_s3_access_key_id)
+{
+    SETTINGS_SET_STRING(settings, s3_access_key_id, bytes_of_s3_access_key_id);
+}
+
+ZarrError
+ZarrStreamSettings_set_s3_secret_access_key(
+  ZarrStreamSettings* settings,
+  const char* s3_secret_access_key,
+  size_t bytes_of_s3_secret_access_key)
```
shlomnissan, line 184: Nit: Are we using clang-tidy for code formatting? If we don't, we should probably configure it for consistency.

src/internal/stream.settings.cpp
```diff
@@ -0,0 +1,539 @@
...
+                  index,
+                  settings->dimensions.size());
+        return ZarrError_InvalidIndex;
+    }
+
+    std::string dim_name = trim(name, bytes_of_name);
+    if (dim_name.empty()) {
+        LOG_ERROR("Invalid name. Must not be empty");
+        return ZarrError_InvalidArgument;
+    }
```
shlomnissan, line 359: - You already checked that `name` is not null, and that `bytes_of_name` is greater than zero. Under what circumstances would `dim_name` be empty?
- If such a case exists, it would be better to check it against a `std::string_view` to avoid a premature allocation of `std::string`.

src/internal/stream.settings.cpp
```diff
@@ -0,0 +1,539 @@
...
+{
+    if (!settings) {
+        LOG_WARNING("Null pointer: settings. Returning 0.");
+        return 0;
+    }
+    return settings->dimensions.size();
+}
+
+ZarrError
+ZarrStreamSettings_get_dimension(const ZarrStreamSettings* settings,
```
shlomnissan, line 481: - I think there needs to be a distinction in the name between an accessor function that returns a value and a function that is used to assign values into output parameters.
- The signature of this function reinforces the need for a struct that represents a "dimension", instead of passing individual properties.

src/internal/stream.settings.cpp
```diff
@@ -0,0 +1,539 @@
...
+                                 size_t* array_size_px,
+                                 size_t* chunk_size_px,
+                                 size_t* shard_size_chunks)
+{
+    EXPECT_VALID_ARGUMENT(settings, "Null pointer: settings");
+    EXPECT_VALID_ARGUMENT(name, "Null pointer: name");
+    EXPECT_VALID_ARGUMENT(kind, "Null pointer: kind");
+    EXPECT_VALID_ARGUMENT(array_size_px, "Null pointer: array_size_px");
+    EXPECT_VALID_ARGUMENT(chunk_size_px, "Null pointer: chunk_size_px");
+    EXPECT_VALID_ARGUMENT(shard_size_chunks, "Null pointer: shard_size_chunks");
```
shlomnissan, line 495: Why would an accessor or query function need to validate its input?

src/internal/stream.settings.cpp
```diff
@@ -0,0 +1,539 @@
...
+{
+    EXPECT_VALID_ARGUMENT(settings, "Null pointer: settings");
+
+    settings->multiscale = multiscale > 0;
+    return ZarrError_Success;
+}
+
+/* Getters */
+const char*
+ZarrStreamSettings_get_store_path(const ZarrStreamSettings* settings)
```
shlomnissan, line 384: I find all the `_get_` functions confusing. I suspect they're unnecessary boilerplate, but I might be overlooking something. My reasoning is simple: if I already own the settings object, why do I need a public API for accessing individual values?

shlomnissan (commented):
Another partial review focusing on `stream.settings`.

src/CMakeLists.txt
```diff
@@ -1,9 +1,54 @@
...
+        blosc_static
+        miniocpp::miniocpp
+)
+
+set_target_properties(${tgt} PROPERTIES
+        MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>"
+)
+
+install(TARGETS ${tgt}
+        LIBRARY DESTINATION lib
```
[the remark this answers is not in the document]
aliddell: Can you explain what you mean by that?

src/include/zarr.h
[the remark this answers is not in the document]
shlomnissan: In my opinion, documentation tagging is an "all or nothing" proposition. Even when a parameter seems obvious, explicitly documenting it maintains consistency and ensures clear, complete documentation. Documentation generation tools use `@params` tags to create detailed references and relationships and IDEs leverage these tags to provide quick references.

src/include/zarr.h
```diff
@@ -0,0 +1,550 @@
...
+     * can be converted to a human-readable error message using
+     * Zarr_get_error_message(). If the function fails, the settings struct is
+     * not modified.
+     **************************************************************************/
+
+    /**
+     * @brief Set the store path for the Zarr stream.
+     * @detail This parameter is required for all Zarr streams. When acquiring
+     * to the filesystem, the store path is a directory path. When acquiring to
+     * S3, the store path is a key prefix.
```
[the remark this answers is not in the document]
aliddell: No, it's not a URL. For an S3 store it'd end up looking something like `$endpoint/$bucket_name/$store_path`. That's why I said "key prefix" here.

aliddell:
@shlomnissan I'm picking this up in #293 and beyond.

PR #293, opened 2024-09-17
Standalone 1/N: Move driver source and test files to separate directories

aliddell:
Splitting up #291. Moves files and updates CMakeLists.txt files accordingly.


src/driver/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
+option(BUILD_ACQUIRE_DRIVER_ZARR "Build the Acquire Zarr driver" ON)
```
shlomnissan, line 1: Nit: It makes more sense to set this option in the root CMakeLists file and conditionally add the driver's subdirectory, rather than wrapping this entire configuration file in an if statement.

tests/CMakeLists.txt
```diff
@@ -1,90 +1,5 @@
...
-                $<TARGET_FILE_DIR:${project}-${onename}>
-                DEPENDS ${driver}
-                COMMENT "Copying ${driver} to $<TARGET_FILE_DIR:${project}-${onename}>"
-        )
-
-        foreach (name ${tests})
-            add_dependencies(${tgt} ${project}-copy-${driver}-for-tests)
-        endforeach ()
-    endforeach ()
+    add_subdirectory(driver)
```
shlomnissan, line 4: Rather than using a separate top-level CMake file for tests, you could simply include this subdirectory in the driver's CMake file.

tests/driver/CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
+if (BUILD_ACQUIRE_DRIVER_ZARR)
```
shlomnissan, line 1: This relates to the previous comment. If you add this subdirectory from the driver's CMake file (which is already added based on the same condition), you could remove this if-statement.

tests/driver/CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+
+    #
+    # PARAMETERS
+    #
+    set(project acquire-driver-zarr) # CMAKE_PROJECT_NAME gets overridden if this is a subtree of another project
+
+    #
+    # Tests
+    #
+    set(tests
```
shlomnissan, line 14: Note: While I believe it's important to manually list all source and header files in the project's main build configuration, I prefer using `file(GLOB TEST_SOURCES ${CMAKE_CURRENT_LIST_DIR}/**/*.cpp)` for test-related builds. This dynamic file loading reduces overhead for test configurations.

This is just a suggestion—feel free to disregard. However, I'd be interested in your thoughts if you disagree.

shlomnissan (approved):
I've made a few minor suggestions to improve the CMake configuration. While not blocking, I believe these changes are worth implementing.

src/CMakeLists.txt
```diff
@@ -1,57 +1,3 @@
-if (NOT TARGET acquire-core-logger)
-    add_subdirectory(../acquire-common/acquire-core-libs ${CMAKE_CURRENT_BINARY_DIR}/acquire-core-libs)
-endif ()
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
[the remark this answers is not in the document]
aliddell: I can save this for the streaming library.

src/driver/CMakeLists.txt
aliddell: Do you mean the top level or the one in src/?

tests/driver/CMakeLists.txt
aliddell: That's an interesting suggestion, but I think I'll save it for later experimentation.

PR #294, opened 2024-09-17
Standalone 2/N: Define the Zarr streaming API

aliddell:
Defines but does not implement the API. Depends on #293.


include/acquire.zarr.h
```diff
@@ -0,0 +1,274 @@
...
+     * @param[in] settings The Zarr stream settings struct.
+     */
+    void ZarrStreamSettings_destroy(ZarrStreamSettings* settings);
+
+    /**
+     * @brief Copy a Zarr stream settings struct.
+     * @param[in] settings The Zarr stream settings struct to copy.
+     * @return A copy of the Zarr stream settings struct.
+     */
+    ZarrStreamSettings* ZarrStreamSettings_copy(
```
shlomnissan, line 59: Is there a compelling reason to implement copying functionality for this object?

include/acquire.zarr.h
```diff
@@ -0,0 +1,274 @@
...
+     * @param[in, out] settings
+     * @param[in] store_path The store path for the Zarr stream. Directory path
+     * when acquiring to the filesystem, key prefix when acquiring to S3.
+     * @param[in] bytes_of_store_path The length of @p store_path in bytes,
+     * including the null terminator.
+     * @param[in] s3_settings Optional S3 settings. If NULL, the store path is
+     * assumed to be a directory path.
+     * @return ZarrStatus_Success on success, or an error code on failure.
+     */
+    ZarrStatus ZarrStreamSettings_set_store(ZarrStreamSettings* settings,
```
shlomnissan, line 73: Since there are two distinct store types (filesystem and S3), I recommend offering separate initialization functions for each type. This approach not only simplifies the API but also makes the intended usage clearer to the user.

Having dedicated functions allows for more specific parameter requirements and error handling tailored to each store type, which can enhance the overall user experience. It reduces ambiguity and helps prevent misuse, ultimately leading to a more intuitive and maintainable interface.

Another way to enhance the API would be to separate the store parameters object from the settings object.

jeskesen:
(Paraphrasing a verbal conversation, please correct if I misremember or misinterpreted):
> @shlomnissan : "Since you allocate the settings structs and pass them in, the user already has the data and the getters are unnecessary, and should thus be removed."
> @jeskesen : "In my previous C++/Python API, I needed to have a getter and setter for each member for pybind11 to create attributes" (in retrospect, not relevant to the conversation)
> @aliddell : "The getters are needed to fulfill the contracts of the Acquire driver API"
 
Do the acquire driver's needs necessarily need to be fulfilled by the public API? Can the getters be exposed for Acquire but not in the standalone implementation?  If this were C++, I'd have said maybe there could be an Acquire friend class or something, but since its C maybe its just 2 different header files (internal & external) or #ifdef inclusion in the internal acquire API?

include/acquire.zarr.h
```diff
@@ -0,0 +1,274 @@
...
+      const ZarrDimensionProperties* dimension);
+
+    /**
+     * @brief Set the multiscale flag for the Zarr stream.
+     * @param[in, out] settings The Zarr stream settings struct.
+     * @param[in] multiscale A flag indicating whether to stream to multiple
+     * levels of detail.
+     * @return ZarrStatus_Success on success, or an error code on failure.
+     */
+    ZarrStatus ZarrStreamSettings_set_multiscale(ZarrStreamSettings* settings,
```
shlomnissan, line 136: The most significant design change we should consider is the management of the settings object ownership. Instead of the API owning this object (allocating space, providing copy and dispose functions, etc.), we should have users instantiate and configure the fields themselves. This approach would eliminate the need for object lifetime management, and mutator and accessor functions. When creating a stream, users would pass a pointer to the settings object (and potentially a pointer to the store object), and the API would internally validate the settings and populate defaults as needed.

include/acquire.zarr.h
```diff
@@ -0,0 +1,274 @@
...
+      ZarrStreamSettings* settings,
+      const char* external_metadata,
+      size_t bytes_of_external_metadata);
+
+    /**
+     * @brief Get the store path for the Zarr stream.
+     * @param settings The Zarr stream settings struct.
+     * @return The store path for the Zarr stream, or NULL on failure.
+     */
+    const char* ZarrStreamSettings_get_store_path(
```
shlomnissan, line 159: Even if we decide to have the public API own the settings object, I don't think we should provide mutators—though an argument could be made for them. More importantly, I see no clear reason for accessor functions. This approach would require converting the settings struct to a C++ class.

include/acquire.zarr.h
```diff
@@ -0,0 +1,274 @@
...
+     * @return The JSON-formatted custom metadata for the Zarr stream, or NULL on
+     * failure.
+     */
+    const char* ZarrStreamSettings_get_custom_metadata(
+      const ZarrStreamSettings* settings);
+
+    /**
+     * @brief Create a Zarr stream.
+     * @param[in, out] settings The settings for the Zarr stream.
+     * @param[in] version The version of the Zarr stream. 2 or 3.
```
shlomnissan, line 227: Why isn't the version included in the settings object?

include/acquire.zarr.h
```diff
@@ -0,0 +1,274 @@
...
+     * @return The version of the Zarr stream.
+     */
+    ZarrVersion ZarrStream_get_version(const ZarrStream* stream);
+
+    /**
+     * @brief Get a copy of the settings for the Zarr stream.
+     * @param stream The Zarr stream struct.
+     * @return A copy of the settings for the Zarr stream.
+     */
+    ZarrStreamSettings* ZarrStream_get_settings(const ZarrStream* stream);
```
shlomnissan, line 268: This relates to my earlier question about `ZarrStreamSettings_copy`. Is there a specific use case for copying stream settings objects?

include/zarr.types.h
```diff
@@ -0,0 +1,133 @@
...
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+    typedef enum
+    {
+        ZarrStatus_Success = 0,
```
shlomnissan, line 15: In some instances, you assign a default value of zero, while in others, you don't. Is there a specific reason for this inconsistency?

include/acquire.zarr.h
```diff
@@ -0,0 +1,274 @@
...
+    const char* ZarrStreamSettings_get_store_path(
+      const ZarrStreamSettings* settings);
+
+    /**
+     * @brief Get the S3 settings for the Zarr stream.
+     * @param settings The Zarr stream settings struct.
+     * @return The S3 settings for the Zarr stream, or an uninitialized struct on
+     * failure.
+     */
+    ZarrS3Settings ZarrStreamSettings_get_s3_settings(
```
shlomnissan, line 168: This illustrates why we should eliminate accessors, even if we opt for the public API to manage the settings object and configure it through accessor functions. If I've already provided S3 store keys, why would I need to retrieve them? Any scenario requiring this retrieval could potentially be considered a security risk.

shlomnissan (commented):
Despite the relatively small size of this PR, I believe it's the most critical as it represents the library's public interface. My main concerns with this API design are:

- Ownership management for the settings object: IMO users should be responsible for managing the object's memory and data. This approach would eliminate all lifetime-related functions (create, destroy, copy, etc.). This will streamline the API by reducing its surface area which is critical for good API design: maximum flexibility with minimum surface area to reduce potential errors and unintended usage.
- A significant improvement would be converting this C++ struct to a C struct and expecting users to fill out the fields before creating a stream. The "create stream" function could then handle data validation, default values, and errors. This change would, I believe, eliminate all mutators and accessors, which feel redundant. This is particularly true for getters, which not only seem unnecessary but may also introduce security risks depending on usage patterns.

include/acquire.zarr.h
```diff
@@ -0,0 +1,274 @@
...
+     * @param status The status code.
+     * @return A human-readable status message.
+     */
+    const char* Zarr_get_error_message(ZarrStatus status);
+
+    /**
+     * @brief Create a Zarr stream settings struct.
+     * @return A pointer to the Zarr stream settings struct, or NULL on failure.
+     */
+    ZarrStreamSettings* ZarrStreamSettings_create();
```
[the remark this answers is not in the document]
jeskesen: To me, I feel like this is the C-compliant way of having constructors (also: destructors, copy constructors, etc).... especially if it does something to initialize data inside the function.  To me, it feels safer to have the data "own" its internals, or at least have some responsibility for them. but I acknowledge that I am very biased towards OOP.

include/zarr.types.h
aliddell: These are all collected from different places, I just moved them and didn't really look at it. ZarrVersion_2 defaults to 2 to eliminate any surprises.

include/acquire.zarr.h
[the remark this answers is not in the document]
aliddell: To answer your second question, functions beginning with `Zarr_` aren't tied either to the stream or settings structs, as you've noticed, but point to API internals (like the logger or the API version).

include/acquire.zarr.h
```diff
@@ -0,0 +1,274 @@
...
+#endif
+
+    typedef struct ZarrStreamSettings_s ZarrStreamSettings;
+    typedef struct ZarrStream_s ZarrStream;
+
+    /**
+     * @brief Get the version of the Zarr API.
+     * @return The version of the Zarr API.
+     */
+    uint32_t Zarr_get_api_version();
```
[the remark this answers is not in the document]
aliddell: I have it defined downstream as a constant macro in the file that implements the function, but I could see it being defined as a constant macro with `target_compile_definitions`. It's a single number, so it's not semver, but we could do some bitmasking to get semver from it.

include/acquire.zarr.h
aliddell: Yes. If the settings are validated when the stream is created, and if the user retains the original data, then the stream should make a copy of the settings struct (or at least the data therein) so that the user can't inadvertently invalidate the stream by messing with the settings.

include/acquire.zarr.h
```diff
@@ -0,0 +1,112 @@
...
+     * @note The store path can be a filesystem path or an S3 key prefix. For example,
+     * supplying an endpoint "s3://my-endpoint.com" and a bucket "my-bucket" with a
+     * store_path of "my-dataset.zarr" will result in the store being written to
+     * "s3://my-endpoint.com/my-bucket/my-dataset.zarr".
+     * @note The dimensions array may be allocated with ZarrStreamSettings_create_dimension_array
+     * and freed with ZarrStreamSettings_destroy_dimension_array. The order in which you
+     * set the dimension properties in the array should match the order of the dimensions
+     * from slowest to fastest changing, for example, [Z, Y, X] for a 3D dataset.
+     */
+    typedef struct ZarrStreamSettings_s
```
shlomnissan, line 25: Is there a specific reason why this type is not included in the `zarr.types.h` file?

include/acquire.zarr.h
```diff
@@ -0,0 +1,112 @@
...
+     * @return The version of the Zarr API.
+     */
+    uint32_t Zarr_get_api_version();
+
+    /**
+     * @brief Set the log level for the Zarr API.
+     * @param level The log level.
+     * @return ZarrStatusCode_Success on success, or an error code on failure.
+     */
+    ZarrStatusCode Zarr_set_log_level(ZarrLogLevel level);
```
shlomnissan, line 51: I left questions about the log API that were not answered in the previous pass. While not critical, I believe there *might* be potential for improvement in this area.

shlomnissan (approved):
Great job 👏 The API surface is significantly reduced, and users face less boilerplate code. The stream creation process is streamlined: create a settings object instance, populate it, and create a stream. This change is excellent!

include/acquire.zarr.h
```diff
@@ -0,0 +1,274 @@
...
+     * @return The version of the Zarr API.
+     */
+    uint32_t Zarr_get_api_version();
+
+    /**
+     * @brief Set the log level for the Zarr API.
+     * @param level The log level.
+     * @return ZarrStatus_Success on success, or an error code on failure.
+     */
+    ZarrStatus Zarr_set_log_level(ZarrLogLevel level);
```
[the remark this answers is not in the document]
aliddell: Yeah, it's the former. I like it here because users can still set to debug and apply any additional filters they like to the log, or they can use the built-in filters. It's not called out in the doc (I can change that), but in the implementation all calculation is skipped for lower-level logs, and this can be a performance improvement.

PR #297, opened 2024-09-17
Standalone 5/N: Implement and test ZarrStream_s

aliddell:
Depends on #295


src/streaming/zarr.stream.cpp
```diff
@@ -0,0 +1,494 @@
...
+    return nullptr != settings->compression_settings;
+}
+
+[[nodiscard]]
+std::string
+trim(const char* s)
+{
+    if (!s || !*s) {
+        return {};
+    }
```
shlomnissan, line 30:
```suggestion
    if (s == nullptr || *s == '\0') {
        return {};
    }
```

src/streaming/zarr.stream.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+bool
+is_compressed_acquisition(const struct ZarrStreamSettings_s* settings)
+{
+    return nullptr != settings->compression_settings;
+}
+
+[[nodiscard]]
+std::string
+trim(const char* s)
```
shlomnissan, line 26: This utility function, unrelated to zarr streaming, can be moved to a "utilities" header file. It can be used to clean up input in various parts of the codebase.

src/streaming/zarr.stream.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+[[nodiscard]]
+bool
+validate_s3_settings(const ZarrS3Settings* settings)
+{
+    std::string trimmed = trim(settings->endpoint);
+    if (trimmed.empty()) {
+        LOG_ERROR("S3 endpoint is empty");
+        return false;
+    }
```
shlomnissan, line 59: The emptiness check is repeated multiple times. We should extract it into a separate function that takes an untrimmed string and an error message as parameters.
```cpp
if (is_empty_string(settings->endpoint, "S3 endpoint is empty")) {
    return false;
}
```
This is also a good candidate for a utility function.

tests/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 if (${NOTEST})
     message(STATUS "Skipping test targets")
 else ()
+    add_subdirectory(unit-tests)
```
shlomnissan, line 4: What other tests are we expecting to necessitate the need for a subdirectory called "unit-tests"?

tests/unit-tests/unit.test.macros.hh
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include "logger.hh"
+
+#define EXPECT(e, ...)                                                         \
```
shlomnissan, line 5: Why don't we use a testing framework like GTest to avoid this boilerplate code every time we configure tests?

src/streaming/zarr.stream.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+#include <filesystem>
+
+namespace fs = std::filesystem;
+
+namespace {
+bool
+is_s3_acquisition(const struct ZarrStreamSettings_s* settings)
+{
+    return nullptr != settings->s3_settings;
```
shlomnissan, line 15: I already mentioned this over Slack. Yoda conditions are uncommon in modern C++ codebases because the compiler and static analyzer will warn you about this. Personally, I find this form less readable and counterintuitive, as it contradicts the natural reading order.

shlomnissan (approved):
I believe there's room for improvement in this PR. My main concern is the new API for creating a dimension array using the settings object for assignment and deallocation. While I'm approving this PR to avoid blocking progress, I think we should consider this comment before merging. It's a borderline issue, but worth considering.

include/acquire.zarr.h
```diff
@@ -7,17 +7,32 @@ extern "C"
...
+     * @note The store path can be a filesystem path or an S3 key prefix. For example,
+     * supplying an endpoint "s3://my-endpoint.com" and a bucket "my-bucket" with a
+     * store_path of "my-dataset.zarr" will result in the store being written to
+     * "s3://my-endpoint.com/my-bucket/my-dataset.zarr".
+     * @note The dimensions array may be allocated with ZarrStreamSettings_create_dimension_array
+     * and freed with ZarrStreamSettings_destroy_dimension_array. The order in which you
+     * set the dimension properties in the array should match the order of the dimensions
+     * from slowest to fastest changing, for example, [Z, Y, X] for a 3D dataset.
+     */
     typedef struct ZarrStreamSettings_s
```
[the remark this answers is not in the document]
aliddell: I didn't want to bury the main settings struct at the bottom of zarr.types.h.

src/streaming/zarr.stream.cpp
```diff
@@ -0,0 +1,494 @@
...
+#include "acquire.zarr.h"
+
+#include <blosc.h>
+
+#include <filesystem>
+
+namespace fs = std::filesystem;
+
+namespace {
+bool
```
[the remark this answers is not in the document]
aliddell: My understanding is that using anonymous namespaces is generally preferred over `static` for internal linkage. Indeed, if I try `static bool` here, clang-tidy complains that it's redundant.

src/streaming/zarr.stream.cpp
aliddell: It's not really used anywhere else, but I can move it to zarr.common.hh/cpp in the next PR

include/acquire.zarr.h
```diff
@@ -48,14 +63,26 @@ extern "C"
...
+     * @param dimension_count The number of dimensions in the dataset to allocate memory for.
+     * @return ZarrStatusCode_Success on success, or an error code on failure.
+     */
+    ZarrStatusCode ZarrStreamSettings_create_dimension_array(ZarrStreamSettings* settings, size_t dimension_count);
+
+    /**
+     * @brief Free memory for the dimension array in the Zarr stream settings struct.
+     * @param[in, out] settings The Zarr stream settings struct containing the dimension array to free.
+     */
+    void ZarrStreamSettings_destroy_dimension_array(ZarrStreamSettings* settings);
```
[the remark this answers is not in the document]
aliddell: It could, but there are two parameters that are set or reset (I double checked this and realized `dimension_count` is not actually getting reset in `ZarrStreamSettings_destroy_dimension_array` though, but I'll push a fix for that).

PR #299, opened 2024-09-17
Standalone 7/N: Implement and test Zarr common functions

aliddell:
Depends on #298.


src/streaming/zarr.common.cpp
```diff
@@ -0,0 +1,297 @@
...
+        case ZarrDataType_int32:
+        case ZarrDataType_uint32:
+        case ZarrDataType_float32:
+            return 4;
+        case ZarrDataType_int64:
+        case ZarrDataType_uint64:
+        case ZarrDataType_float64:
+            return 8;
+        default:
+            throw std::runtime_error("Invalid data type: " +
```
shlomnissan, line 25: How is this exception handled? Is it caught at a higher level and returned as an error code to the C interface?

src/streaming/zarr.common.cpp
```diff
@@ -0,0 +1,297 @@
...
+                                     std::to_string(data_type));
+    }
+}
+
+size_t
+zarr::bytes_of_frame(const std::vector<Dimension>& dims, ZarrDataType type)
+{
+    return bytes_of_type(type) * dims.back().array_size_px *
+           dims[dims.size() - 2].array_size_px;
```
shlomnissan, line 34: What do the last and second-to-last elements represent? Assigning these values to named variables could enhance clarity. For instance, if they represent width and height, you might rewrite it as:
```cpp
const auto width = dims[dims.size() - 1].array_size_px;
const auto height = dims[dims.size() - 2].array_size_px;
return bytes_of_type(type) * width * height;
```

src/streaming/zarr.common.cpp
```diff
@@ -0,0 +1,297 @@
...
+           dimension.chunk_size_px;
+}
+
+size_t
+zarr::shards_along_dimension(const Dimension& dimension)
+{
+    const size_t shard_size = dimension.shard_size_chunks;
+    if (shard_size == 0) {
+        return 0;
+    }
```
shlomnissan, line 52: Nit: Add a check for zero value before the assignment to prevent an unnecessary copy.
```cpp
if (dimension.shard_size_chunks == 0) {
    return 0;
}
const auto shard_size = dimension.shard_size_chunks;
```

src/streaming/zarr.common.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+    const size_t n_chunks = chunks_along_dimension(dimension);
```
shlomnissan, line 54: This is the most compelling set of arguments I've encountered for favoring `auto` over explicitly named types.

https://learning.oreilly.com/library/view/effective-modern-c/9781491908419/ch02.html

src/streaming/zarr.common.cpp
```diff
@@ -0,0 +1,297 @@
...
+zarr::chunk_lattice_index(size_t frame_id,
+                          size_t dimension_idx,
+                          const std::vector<Dimension>& dims)
+{
+    // the last two dimensions are special cases
+    EXPECT(dimension_idx < dims.size() - 2,
+           "Invalid dimension index: %zu",
+           dimension_idx);
+
+    // the first dimension is a special case
```
shlomnissan, line 68: To improve safety and maintainability when dealing with special cases in our dimension handling, we could consider creating a dedicated type that provides a robust interface for accessing dimension data. This approach would allow us to encapsulate the logic associated with special cases, giving them meaningful names that convey their purpose. By doing so, we can provide bounded access to the non-special case dimension data, reducing the risk of errors and enhancing code readability. Such an interface would promote clearer usage patterns and facilitate easier updates in the future, as any changes to special case handling would be centralized within this type, rather than scattered throughout the codebase.

I think this is important since this knowledge is spread across multiple functions.

src/streaming/zarr.common.cpp
```diff
@@ -0,0 +1,297 @@
...
+size_t
+zarr::tile_group_offset(size_t frame_id, const std::vector<Dimension>& dims)
+{
+    std::vector<size_t> strides;
+    strides.push_back(1);
+    for (auto i = dims.size() - 1; i > 0; --i) {
+        const auto& dim = dims[i];
+        CHECK(dim.chunk_size_px);
+        const auto a = dim.array_size_px, c = dim.chunk_size_px;
+        strides.insert(strides.begin(), strides.front() * ((a + c - 1) / c));
```
shlomnissan, line 103: Using `strides.insert(strides.begin(), ...)` in a loop may not be the most efficient approach since inserting at the beginning of a vector requires shifting all elements. It might be more efficient to reserve space ahead of time and fill the vector from the end to the beginning.

shlomnissan (approved):
LGTM!

src/streaming/zarr.common.cpp
aliddell: Yeah, I think this is a fine idea.

PR #300, opened 2024-09-17
Standalone 8/N: Implement and test ThreadPool, S3Connection{Pool}.

aliddell:
Depends on #299.


src/streaming/thread.pool.hh
```diff
@@ -0,0 +1,56 @@
...
+    ~ThreadPool() noexcept;
+
+    /**
+     * @brief Push a job onto the job queue.
+     *
+     * @param job The job to push onto the queue.
+     * @return true if the job was successfully pushed onto the queue, false
+     * otherwise.
+     */
+    [[nodiscard]] bool push_to_job_queue(JobT&& job);
```
shlomnissan, line 33: I suggest renaming it to `push_job` or `add_job`. There's no need to expose implementation details.

src/streaming/thread.pool.hh
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * @brief Block until all jobs on the queue have processed, then spin down
+     * the threads.
+     * @note After calling this function, the job queue no longer accepts jobs.
+     */
+    void await_stop() noexcept;
+
+  private:
+    std::function<void(const std::string&)> error_handler_;
```
shlomnissan, line 43: Ah, if this signature is defined twice, an alias will definitely be useful.

src/streaming/thread.pool.hh
```diff
@@ -0,0 +1,56 @@
...
+
+    std::vector<std::thread> threads_;
+    mutable std::mutex jobs_mutex_;
```
shlomnissan, line 46: Same question about why `mutable` is needed in this case.

src/streaming/thread.pool.hh
```diff
@@ -0,0 +1,56 @@
...
+    std::condition_variable cv_;
+    std::queue<JobT> jobs_;
+
+    std::atomic<bool> is_accepting_jobs_;
```
shlomnissan, line 50: The same comment applies regarding the preference for inline initialization over the constructor's initializer list.

src/streaming/thread.pool.hh
```diff
@@ -0,0 +1,56 @@
...
+
+    std::optional<ThreadPool::JobT> pop_from_job_queue_() noexcept;
+    [[nodiscard]] bool should_stop_() const noexcept;
+    void thread_worker_();
```
shlomnissan, line 54: This function name is confusing. I suggest renaming it to `process_jobs()` if that accurately reflects its implementation.

src/streaming/thread.pool.cpp
```diff
@@ -0,0 +1,96 @@
+#include "thread.pool.hh"
+
+zarr::ThreadPool::ThreadPool(unsigned int n_threads,
+                             std::function<void(const std::string&)> err)
+  : error_handler_{ err }
```
shlomnissan, line 5: `std::function` can be costly in terms of performance. Moreover, it's common practice to use inline lambdas for scenarios like error callbacks. I recommend shifting to function objects to address these situations more efficiently.

```cpp
: error_handler_ { std::move(err) }
```

src/streaming/thread.pool.cpp
```diff
@@ -0,0 +1,96 @@
...
+  , is_accepting_jobs_{ true }
+{
+    n_threads = std::clamp(
+      n_threads, 1u, std::max(std::thread::hardware_concurrency(), 1u));
```
shlomnissan, line 9: A clearer way to write this might be:

```cpp
const auto max_threads = std::max(std::thread::hardware_concurrency(), 1u);
n_threads = std::clamp(n_threads, 1u, max_threads);
```

src/streaming/thread.pool.cpp
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+
+    await_stop();
+}
+
+bool
+zarr::ThreadPool::push_to_job_queue(JobT&& job)
+{
+    std::unique_lock lock(jobs_mutex_);
```
shlomnissan, line 31: A similar question arises about the need to mix different locking mechanisms. I'm not an expert when it comes to the locking API, so there might be a compelling reason for this approach that I'm overlooking.

shlomnissan (approved):
I have no blocking comments here. However, there is some room for improvement.

src/streaming/s3.connection.hh
```diff
@@ -0,0 +1,141 @@
...
+#include <mutex>
+#include <string>
+#include <span>
+#include <vector>
+
+namespace zarr {
+class S3Connection
+{
+  public:
+    explicit S3Connection(const std::string& endpoint,
```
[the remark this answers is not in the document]
aliddell: `explicit` here is probably vestigial

src/streaming/s3.connection.hh
```diff
@@ -0,0 +1,141 @@
...
+                          const std::string& access_key_id,
```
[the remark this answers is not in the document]
aliddell: You end up needing to create `std::string`s in the constructor from these, because `minio::s3::BaseUrl` and `minio::creds::StaticProvider` won't take `std::string_view` and it appears there's no implicit conversion.

src/streaming/s3.connection.hh
```diff
@@ -0,0 +1,141 @@
...
+};
+
+class S3ConnectionPool
+{
+  public:
+    S3ConnectionPool(size_t n_connections,
+                     const std::string& endpoint,
+                     const std::string& access_key_id,
+                     const std::string& secret_access_key);
+    ~S3ConnectionPool() noexcept;
```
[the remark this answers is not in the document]
aliddell: Destructors are also a good example, no? They're often (usually?) called during stack unwinding, e.g., during exception handling.

src/streaming/s3.connection.hh
```diff
@@ -0,0 +1,141 @@
...
+
+    std::unique_ptr<S3Connection> get_connection();
+    void return_connection(std::unique_ptr<S3Connection>&& conn);
+
+  private:
+    std::vector<std::unique_ptr<S3Connection>> connections_;
+    mutable std::mutex connections_mutex_;
```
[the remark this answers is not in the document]
aliddell: Vestigial.

src/streaming/s3.connection.cpp
```diff
@@ -0,0 +1,254 @@
...
+{
+    EXPECT(!bucket_name.empty(), "Bucket name must not be empty.");
+    EXPECT(!object_name.empty(), "Object name must not be empty.");
+
+    minio::s3::StatObjectArgs args;
+    args.bucket = bucket_name;
+    args.object = object_name;
+
+    auto response = client_->StatObject(args);
+    // casts to true if response code in 200 range and error message is empty
```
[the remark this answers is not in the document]
aliddell: I don't like the idea of digging into the members of the Response struct and maintaining an explicit check on those members, which may change in the future, given the bool cast is already what this is intended for.

src/streaming/s3.connection.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+    LOG_DEBUG("Creating multipart object %s in bucket %s",
+              object_name.data(),
+              bucket_name.data());
+    minio::s3::CreateMultipartUploadArgs args;
+    args.bucket = bucket_name;
+    args.object = object_name;
+
+    auto response = client_->CreateMultipartUpload(args);
+    CHECK(!response.upload_id.empty());
```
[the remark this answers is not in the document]
aliddell: Roughly, I treat a `CHECK` a bit like an assert. (I'm sure you can find examples of my not being consistent with this.) But more logging is ~~always~~ usually better.

src/streaming/s3.connection.cpp
```diff
@@ -0,0 +1,254 @@
...
+    EXPECT(!object_name.empty(), "Object name must not be empty.");
+    EXPECT(!data.empty(), "Number of bytes must be positive.");
+    EXPECT(part_number, "Part number must be positive.");
+
+    LOG_DEBUG("Uploading multipart object part %zu for object %s in bucket %s",
+              part_number,
+              object_name.data(),
+              bucket_name.data());
+
+    std::string_view sv(reinterpret_cast<const char*>(data.data()),
```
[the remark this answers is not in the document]
aliddell: Not unless there's an implicit conversion. `UploadPartArgs` wants `string_view` in that field.

src/streaming/thread.pool.hh
```diff
@@ -0,0 +1,56 @@
...
+#include <optional>
+#include <queue>
+#include <string>
+#include <thread>
+
+namespace zarr {
+class ThreadPool
+{
+  public:
+    using JobT = std::function<bool(std::string&)>;
```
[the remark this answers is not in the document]
aliddell: The `Task` (I'm going to call it a task, I like that one) is not strictly speaking a callback, but a closure which both returns false and sets an error message on failure (which is why the `std::string&` argument is not const).

src/streaming/thread.pool.hh
aliddell: I think also vestigial.

src/streaming/thread.pool.cpp
aliddell: Something like this?

```c++
class ErrorCallerbacker {
public:
    ErrorCallbacker(std::unique_ptr<ZarrStream_s> stream) : stream_(stream) {}
    int operator()(const std::string& s) const { stream_->set_error(s); }
private:
    int stream_;
};
```

src/streaming/s3.connection.cpp
```diff
@@ -0,0 +1,254 @@
...
+std::string
+zarr::S3Connection::put_object(std::string_view bucket_name,
+                               std::string_view object_name,
+                               std::span<uint8_t> data)
+{
+    EXPECT(!bucket_name.empty(), "Bucket name must not be empty.");
+    EXPECT(!object_name.empty(), "Object name must not be empty.");
+    EXPECT(!data.empty(), "Data must not be empty.");
+
+    minio::utils::CharBuffer buffer((char*)const_cast<uint8_t*>(data.data()),
```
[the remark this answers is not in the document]
shlomnissan: `const_cast` is considered bad because it can break the safety guarantees provided by const in C++. The const qualifier ensures that a value cannot be modified after it is initialized, which helps avoid accidental changes and makes code more predictable and easier to reason about.

The only reasonable explanation for using it is interfacing with legacy code that doesn't respect const correctness or maybe working with APIs that incorrectly omit const qualifiers in places where they should, but both cases are uncommon.

Why did you use it in the first place?

src/streaming/thread.pool.cpp
shlomnissan: Ignore the last part of my comment. My primary point was moving the object instead of copying it.

PR #302, opened 2024-09-17
Standalone 10/N: Implement and test base ArrayWriter

aliddell:
Depends on #301.


src/streaming/array.writer.hh
```diff
@@ -0,0 +1,96 @@
...
+
+/// @brief Downsample the array writer configuration to a lower resolution.
+/// @param[in] config The original array writer configuration.
+/// @param[out] downsampled_config The downsampled array writer configuration.
+/// @return True if @p downsampled_config can be downsampled further.
+/// This is determined by the chunk size in @p config. This function will return
+/// false if and only if downsampling brings one or more dimensions lower than
+/// the chunk size along that dimension.
+[[nodiscard]] bool
+downsample(const ArrayWriterConfig& config,
```
shlomnissan, line 34: Why is `downsample` implemented as a free function rather than a public member? If it needs to be invoked outside the context of an `ArrayWriter` instance but is still related to it, it could be a static member. I don't think it's significant either way; I'm just trying to understand the reasoning behind this choice.

src/streaming/array.writer.cpp
```diff
@@ -0,0 +1,422 @@
...
+#ifdef min
+#undef min
+#endif
+
+bool
+zarr::downsample(const ArrayWriterConfig& config,
+                 ArrayWriterConfig& downsampled_config)
+{
+    // downsample dimensions
+    std::vector<ZarrDimension> downsampled_dims;
```
shlomnissan, line 21: If you know the size of `downsampled_dims` (I assume it's `config.dimensions->ndims`), you should use `std::array`. If you must use a dynamically sized array, you can allocate space for it ahead of time to avoid unnecessary allocations.

src/streaming/array.writer.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+            downsampled_dims.emplace_back(dim.name,
+                                          dim.type,
+                                          array_size_px,
+                                          chunk_size_px,
+                                          shard_size_chunks);
+        }
+    }
+    downsampled_config.dimensions = std::make_shared<ArrayDimensions>(
+      std::move(downsampled_dims), config.dtype);
```
shlomnissan, line 53: Why are you using a shared pointer here? A unique pointer should suffice since `downsampled_config` is its exclusive owner.

src/streaming/array.writer.cpp
```diff
@@ -0,0 +1,422 @@
...
+  , append_chunk_index_{ 0 }
+  , is_finalizing_{ false }
+{
+}
+
+size_t
+zarr::ArrayWriter::write_frame(const uint8_t* data, size_t nbytes)
+{
+    const size_t nbytes_frame =
+      bytes_of_frame(*config_.dimensions, config_.dtype);
```
shlomnissan, line 97: Suggestion: Use `auto` more consistently. If you need a type conversion, make it explicit by using `static_cast`.

src/streaming/array.writer.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+    if (nbytes_frame != nbytes) {
+        LOG_WARNING("Frame size mismatch: expected %zu, got %zu. Skipping",
```
shlomnissan, line 100: Should this be categorized as a warning or an error?

src/streaming/array.writer.cpp
```diff
@@ -0,0 +1,422 @@
...
+                        std::to_string(append_chunk_index_);
+            break;
+        default:
+            LOG_ERROR("Unsupported Zarr version");
+            return false;
+    }
+
+    SinkCreator creator(thread_pool_, s3_connection_pool_);
+
+    if (config_.bucket_name && !creator.make_data_sinks(*config_.bucket_name,
```
shlomnissan, line 151: Are you using `bucket_name` to determine if an S3 sink should be created? This approach can lead to issues and doesn't enhance readability. I suggest creating a computed property in the configuration object—something like `is_s3_sink`—that encapsulates this logic. This approach allows you to modify the logic in one central location if needed, while also clearly conveying your intent.

src/streaming/array.writer.cpp
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+    return true;
+}
+
+bool
+zarr::ArrayWriter::make_metadata_sink_()
+{
+    if (metadata_sink_) {
+        return true;
```
shlomnissan, line 176: This logic is confusing. If a function named `make_metadata_sink_` returns true, I'd expect it to have created a new metadata sink. However, in this case, it returns true even when a new sink isn't created (if it was created previously). I suggest either logging an error and returning false, or at the very least adding an info log before returning true.

src/streaming/array.writer.cpp
```diff
@@ -0,0 +1,422 @@
...
+        return false;
+    }
+
+    return true;
+}
+
+void
+zarr::ArrayWriter::make_buffers_() noexcept
+{
+    LOG_DEBUG("Creating chunk buffers");
```
shlomnissan, line 216: How do you differentiate between "debug" and "info" log levels?

shlomnissan (approved):
LGTM

src/streaming/array.writer.hh
```diff
@@ -0,0 +1,96 @@
...
+           ArrayWriterConfig& downsampled_config);
+
+class ArrayWriter
+{
+  public:
+    ArrayWriter(const ArrayWriterConfig& config,
+                std::shared_ptr<ThreadPool> thread_pool,
+                std::shared_ptr<S3ConnectionPool> s3_connection_pool);
```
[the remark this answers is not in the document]
aliddell: Can add a second constructor without the connection pool argument

src/streaming/array.writer.hh
aliddell: I think we could probably (in later PRs) directly generate downsampled array writers without the intermediate step of generating configs, and get rid of this entirely.

PR #303, opened 2024-09-17
Standalone 11/N: Implement and test ZarrV2ArrayWriter

aliddell:
Depends on #302.


src/streaming/zarrv2.array.writer.cpp
```diff
@@ -0,0 +1,177 @@
...
+        case ZarrDataType_int32:
+            return dtype_prefix + "i4";
+        case ZarrDataType_int64:
+            return dtype_prefix + "i8";
+        case ZarrDataType_float32:
+            return dtype_prefix + "f4";
+        case ZarrDataType_float64:
+            return dtype_prefix + "f8";
+        default:
+            throw std::runtime_error("Invalid data type: " +
```
shlomnissan, line 41: Are you catching this error somewhere? Why are you using an exception here instead of returning a boolean and logging an error, as you do in the rest of the codebase?

src/streaming/zarrv2.array.writer.cpp
```diff
@@ -0,0 +1,177 @@
...
+  const ArrayWriterConfig& config,
+  std::shared_ptr<ThreadPool> thread_pool,
+  std::shared_ptr<S3ConnectionPool> s3_connection_pool)
+  : ArrayWriter(config, thread_pool, s3_connection_pool)
+{
+}
+
+zarr::ZarrV2ArrayWriter::~ZarrV2ArrayWriter()
+{
+    is_finalizing_ = true;
```
shlomnissan, line 57: I've mentioned this before, but I think that having code that can fail in the destructor is actively bad. Destructors should not fail, throw exceptions, or log errors because there's no way to handle these issues gracefully—you don't know exactly how or when they might be executed. There's also a potential for memory leaks, and no explicit control over resource management. This approach can lead to unpredictable behavior and difficult-to-debug issues.

src/streaming/zarrv2.array.writer.cpp
```diff
@@ -0,0 +1,177 @@
...
+    } catch (...) {
+        LOG_ERROR("Failed to finalize array writer: (unknown)");
+    }
+}
+
+ZarrVersion
+zarr::ZarrV2ArrayWriter::version_() const
+{
+    return ZarrVersion_2;
+}
```
shlomnissan, line 71: I typically implement accessors and mutators in the header file, as they are ideal candidates for inlining.

src/streaming/zarrv2.array.writer.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+bool
+zarr::ZarrV2ArrayWriter::flush_impl_()
+{
+    // create chunk files
+    CHECK(data_sinks_.empty());
+    if (!make_data_sinks_()) {
```
shlomnissan, line 78: Consider adding a log entry here to explain what occurred. This could provide useful context.

src/streaming/zarrv2.array.writer.cpp
```diff
@@ -0,0 +1,177 @@
...
+                                size = chunk.size(),
+                                &latch](std::string& err) -> bool {
+                         bool success = false;
+                         try {
+                             CHECK(sink->write(0, data, size));
+                             success = true;
+                         } catch (const std::exception& exc) {
+                             err = "Failed to write chunk: " +
+                                   std::string(exc.what());
+                         } catch (...) {
```
shlomnissan, line 101: I've mentioned this several times in previous PR reviews: a catch-all exception handler is actively bad and should always be avoided. There are numerous reasons for this, including:

• Loss of specific exception information
• Unclear error handling logic
• Potential for silent failures
• Difficulties in maintaining exception safety guarantees

I strongly advise against using catch-all handlers. Instead, when an unexpected error occurs, it should be handled appropriately and specifically at the time it happens.

shlomnissan (changes requested):
This PR is mostly acceptable. However, I'm requesting changes (for the same reason as in PR #9) because I believe we should remove business logic and logging from the destructor. Instead, these operations should be explicit.

src/streaming/zarrv2.array.writer.cpp
aliddell: It's already logged in the function.

src/streaming/zarrv2.array.writer.cpp
aliddell: Ultimately it bubbles up to `Writer::write_frame()`, and the caller of that would need to catch it. Although I think we can handle it a bit more gracefully.

shlomnissan (approved):
Changes LGTM :+1:

PR #305, opened 2024-09-18
Standalone 13/N: Integrate components into ZarrStream_s

aliddell:
Depends on #304.


src/streaming/array.writer.cpp
```diff
@@ -239,12 +239,12 @@ zarr::ArrayWriter::make_buffers_() noexcept
...
     for (auto& buf : chunk_buffers_) {
         buf.resize(nbytes);
-        std::fill_n(buf.begin(), nbytes, std::byte(0));
+        std::fill(buf.begin(), buf.end(), std::byte(0));
     }
 }
 
 size_t
-zarr::ArrayWriter::write_frame_to_chunks_(std::span<std::byte> data)
+zarr::ArrayWriter::write_frame_to_chunks_(std::span<const std::byte> data)
```
shlomnissan, line 247: Interesting choice to mark the underlying data of the span as const. What's the reasoning behind this?

src/streaming/zarr.stream.cpp
```diff
@@ -414,54 +600,409 @@ ZarrStream_s::create_store_()
...
+        if (scaled_frames_[i]) {
+            average2(dst, bytes_of_data, *scaled_frames_[i], bytes_of_data);
+            std::span frame_data{ reinterpret_cast<const std::byte*>(dst),
+                                  bytes_of_data };
+            EXPECT(writers_[i]->write_frame(frame_data),
+                   "Failed to write frame to writer %zu",
+                   i);
+
+            // clean up this LOD
+            delete[] *scaled_frames_[i];
```
shlomnissan, line 953: Have you considered managing resource lifetimes with smart pointers? I think that modern C++ code should avoid naked new/delete operations, except when dealing with legacy code compatibility or API integration.

shlomnissan (approved):
LGTM :+1:

PR #308, opened 2024-10-24
Add acquire-zarr as a submodule

aliddell:
- Add acquire-zarr submodule
- Make CMake options of informal definitions `NOTEST` and `WITH_EXAMPLES`
  - Conditionally include directories based on option values
- Move src/driver/* to src/
- Move tests/driver/* to tests/
- Remove all streaming code


aliddell:
It looks like a lot, but most things are just deletes and moves.

CMakeLists.txt
```diff
@@ -20,10 +20,15 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
...
 
 option(NOTEST "Disable all tests" OFF)
-option(BUILD_ACQUIRE_DRIVER_ZARR "Build the Acquire Zarr driver" ON)
+option(NO_UNIT_TESTS "Disable unit tests" OFF)
+option(WITH_EXAMPLES "Build examples" OFF)
 
 add_subdirectory(src)
-add_subdirectory(tests)
-add_subdirectory(examples)
+if (NOT NOTEST)
```
jeskesen, line 27: Migrate this to use ENABLE_TESTING?

jeskesen (approved):
Take a look at my comment.  This should probably be done now, to align it with what the acquire-zarr submodule does. 

I'm approving, just to not block it when that (trivial) fix is done.

